PR #5, opened 2021-04-20
Add a simple benchmark for ActiveRecord

tenderlove:
This benchmark just creates a bunch of records in an in-memory database
and then reads the records back out.  The "read" is the part that is
being measured

My machine seems kind of noisy (the stddev is high), but here are the results:

```
ruby 3.1.0dev (2021-04-20T20:54:33Z yjit 1986541a57) [x86_64-darwin20]
YJIT is enabled
last_commit=cYjitCodeComment is only defined if we're not in debugging mode
git branch yjit
git commit hash 1986541a57
-------------  -----------  ----------  ---------  ----------  -----------  -------
bench          interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit  1st itr
active_record  21.1         7.5         19.6       8.8         1.08         0.91   
-------------  -----------  ----------  ---------  ----------  -----------  -------
Legend:
- interp/yjit: ratio of interp/yjit time. Higher is better. Above 1 represents a speedup.
- 1st itr: ratio of interp/yjit time for the first benchmarking iteration.
```

### DIFF
--- a/benchmarks/active_record/benchmark.rb
+++ b/benchmarks/active_record/benchmark.rb
@@ -1,0 +1,45 @@
+require "harness"
+require "bundler/inline"
+require "securerandom"
+
+gemfile do
+  source "https://rubygems.org"
+  gem "activerecord", "~> 6.0.3", ">= 6.0.3.6"
+  gem "sqlite3", "~> 1.4"
+end
+
+require "active_record"
+
+ActiveRecord::Base.establish_connection adapter: "sqlite3", database: ":memory:"
+
+ActiveRecord::Schema.define do
+  create_table :posts, force: true do |t|
+    t.string :title, null: false
+    t.string :body
+    t.string :type_name, null: false
+    t.string :key, null: false
+    t.integer :upvotes, null: false
+    t.integer :author_id, null: false
+    t.timestamps
+  end
+end
+
+class Post < ActiveRecord::Base; end
+
+50000.times {
+  Post.create!(title: SecureRandom.alphanumeric(30),
+               type_name: SecureRandom.alphanumeric(10),
+               key: SecureRandom.alphanumeric(10),
+               body: SecureRandom.alphanumeric(100),
+               upvotes: rand(30),
+               author_id: rand(30))
+}
+
+# heat any caches
+Post.where(id: 1).first.title
+
+run_benchmark(10) do
+  1000.times do |i|
+    Post.where(id: i + 1).first.title
+  end
+end


benchmarks/active_record/benchmark.rb
```diff
@@ -0,0 +1,33 @@
...
+class Post < ActiveRecord::Base; end
+post = nil
+50000.times { post = Post.create!(title: "hi", body: "hello!") }
+
+id = post.id
+# heat any caches
+Post.where(id: id).first.title
+
+run_benchmark(10) do
+  100.times do |i|
```
maximecb, line 30: I would suggest doing 1000 iterations here. A longer-running benchmark will reduce the noise.

benchmarks/active_record/benchmark.rb
```diff
@@ -0,0 +1,33 @@
...
+ActiveRecord::Schema.define do
+  create_table :posts, force: true do |t|
+    t.string :title
+    t.string :body
+  end
+end
+
+class Post < ActiveRecord::Base; end
+post = nil
+50000.times { post = Post.create!(title: "hi", body: "hello!") }
```
maximecb, line 23: Let's bump that to 200K

tenderlove: Does that matter?  We're not fetching all 50k objects in the benchmark.

maximecb:
Cool. You're right that the measurements are very noisy. Too noisy to really conclude anything ATM. Good move trying it on AWS.

I would like to see the running time and complexity increased a bit. Maybe adding more iterations, creating more posts, and adding a few extra fields to the schema (what kind of fields would a realistic web forum have on a post, for instance, maybe upvotes, author_id, etc?).

benchmarks/active_record/benchmark.rb
maximecb: Yes that's a good point. Let's keep your 50K record and fetch more instead, like 1000 or 2000. That will benchmark Ruby more heavily, which is what we want.

tenderlove:
I added more columns and increased the iterations.  Benchmark results on my machine look like this:

```
Total time spent benchmarking: 68s

ruby 3.1.0dev (2021-04-20T21:22:37Z yjit 24128733a7) [x86_64-darwin20]
YJIT is enabled
last_commit=cYjitCodeComment is only defined if we're not in debugging mode
git branch yjit
git commit hash 24128733a7
-------------  -----------  ----------  ---------  ----------  -----------  -------
bench          interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit  1st itr
active_record  218.7        3.4         190.3      4.0         1.15         1.16   
-------------  -----------  ----------  ---------  ----------  -----------  -------
Legend:
- interp/yjit: ratio of interp/yjit time. Higher is better. Above 1 represents a speedup.
- 1st itr: ratio of interp/yjit time for the first benchmarking iteration.
```

maximecb:
Perfect!

PR #18, opened 2021-06-03
Better debugging when running external processes

noahgibbs:
This is a check_call change that makes it a lot clearer what went wrong when something goes wrong. The keyword arguments to check_call ('verbose' and 'debug') can be used at the call site. But their primary purpose is that I can locally default them to true if I want verbose or debug when I'm hunting an error.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -9,10 +9,19 @@ require 'json'
 
 WARMUP_ITRS = 15
 
-def check_call(command)
-    puts(command)
-    status = system(command)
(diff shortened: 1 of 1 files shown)


maximecb:
I agree with the general sentiment but it feels a bit superfluous to add 3 extra arguments.

Maybe just a verbose flag that defaults to false would accomplish the goal?

maximecb:
I think just logging the command when it fails would generally be sufficient. We could log the current working directory as well.

PR #21, opened 2021-06-08
Add jekyll-perf benchmark; have the harness code use absolute paths internally

noahgibbs:
Add a jekyll-perf benchmark based on https://earthly.dev/blog/jruby/ and https://github.com/agbell/jekyll-perf

I've trimmed his example Jekyll app (the Minimal Mistakes theme) to have a lot fewer large images, which doesn't seem to affect the site-build time at all. But I don't think we want them in our Git repo. I've also switched back from Pandoc to Kramdown, which should increase the amount of rendering done in Ruby (he switched to Pandoc because TruffleRuby can't run Kramdown, but it runs via a sub-process.)

Since Jekyll expects to run from a particular directory, I made a minor tweak to the harness code to convert the output CSV path to absolute rather than writing a relative path. That allows a benchmark to change directory without confusing the harness.

On an AWS c5 metal instance, I see a 7% speedup with YJIT:

~~~
yjit_opts=""
ruby_version="ruby 3.1.0dev (2021-06-07T21:30:18Z main a0a7f50e6a) [x86_64-linux] YJIT is enabled"
git_branch="main"
git_commit="a0a7f50e6a"

-----------  -----------  ----------  ---------  ----------  -----------  ------------
bench        interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit  yjit 1st itr
jekyll-perf  9823.3       1.7         9182.5     2.5         1.07         1.07
-----------  -----------  ----------  ---------  ----------  -----------  ------------
~~~


maximecb:
Good work!

Also happy to see that we deliver on this benchmark as well, and even on the first iteration :)

eregon:
> (he switched to Pandoc because TruffleRuby can't run Kramdown, but it runs via a sub-process.)

Just wanted to clarify Kramdown works fine on TruffleRuby, it's just a performance warning due to creating many regexps from a single literal (https://github.com/oracle/truffleruby/issues/2363).

@noahgibbs Thanks for packaging this as a benchmark :)

noahgibbs:
Thrilled to help :-)

PR #23, opened 2021-06-11
Improve the harness and make it more idiomatic

eregon:
When running benchmarks in this repository, I decided to refactor the harness because it felt fairly unidiomatic to me, and it was harder to figure out what it does than needed.
Also, it was requiring the csv stdlib which is unnecessary and loads quite a bit of code as a side effect.

* No need for CSV to print a float per line.
* Use `Benchmark.realtime {}` instead of doing it manually.
* Constants instead of global variables.
* Default the output file to include RUBY_ENGINE and RUBY_ENGINE_VERSION.
* Print RUBY_DESCRIPTION when running the benchmark.
* Include RUBY_DESCRIPTION as the first line in the output.

---

I also made a simple script to summarize the result files in https://github.com/eregon/yjit-bench/blob/truffleruby/analyze.rb
```
$ ruby ../../analyze.rb *.csv
	speedup	median	average	[min   -   max]
ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-linux] - ruby-2.7.3
	1.0x	2.721	2.727	[2.707 - 2.755]
ruby 3.0.1p64 (2021-04-05 revision 0fb782ee38) [x86_64-linux] - ruby-3.0.1
	0.9x	2.992	2.978	[2.948 - 2.997]
```
I was thinking to leave that for another PR though, but I could also add it here, as you prefer.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -225,8 +225,8 @@ def run_benchmarks(ruby_opts, name_filters, out_path)
 
         # Read the benchmark data
         # Convert times to ms
-        rows = CSV.read(ENV["OUT_CSV_PATH"])
-        times = rows[0].map { |v| 1000 * v.to_f }
+        ruby_description, *times = File.readlines(ENV["OUT_CSV_PATH"])
+        times = times.map { |v| 1000 * Float(v) }
         bench_times[bench_name] = times
     end
 


harness/harness.rb
```diff
@@ -1,51 +1,38 @@
...
 # Minimum number of benchmarking iterations
-$MIN_BENCH_ITRS = 10
+MIN_BENCH_ITRS = ENV.fetch('MIN_BENCH_ITRS', 10).to_i
 
 # Minimum benchmarking time in seconds
-$MIN_BENCH_TIME = 10
+MIN_BENCH_TIME = ENV.fetch('MIN_BENCH_TIME', 10).to_i
 
-$out_csv_path = File.expand_path(ENV.fetch('OUT_CSV_PATH', 'output.csv'))
+OUT_CSV_PATH = ENV.fetch('OUT_CSV_PATH', "results-#{RUBY_ENGINE}-#{RUBY_ENGINE_VERSION}.csv")
```
noahgibbs, line 12: If you get rid of the File.expand_path, I think a benchmark that changes the current directory will stop working unless you manually supply an absolute path on the command line.

eregon, line 12: We'd need some other component here to make the default file different between ruby-head and yjit (same for mjit). Not sure what would work best. The full `RUBY_DESCRIPTION` feels a bit weird for a filename.

eregon: Right, I'll restore the `File.expand_path`.

noahgibbs: Here's the RUBY_DESCRIPTION I see currently with my local ruby-yjit:

~~~
irb(main):001:0> RUBY_DESCRIPTION
=> "ruby 3.1.0dev (2021-06-10T17:53:36Z yjit_stats_ruby_de.. 06bd05df65) [x86_64-darwin20]"
~~~

My vote would be *not* to include that in the output filename.

harness/harness.rb
```diff
@@ -1,51 +1,39 @@
...
 # Minimum number of benchmarking iterations
-$MIN_BENCH_ITRS = 10
+MIN_BENCH_ITRS = ENV.fetch('MIN_BENCH_ITRS', 10).to_i
 
 # Minimum benchmarking time in seconds
-$MIN_BENCH_TIME = 10
+MIN_BENCH_TIME = ENV.fetch('MIN_BENCH_TIME', 10).to_i
 
-$out_csv_path = File.expand_path(ENV.fetch('OUT_CSV_PATH', 'output.csv'))
+default_path = "results-#{RUBY_ENGINE}-#{RUBY_ENGINE_VERSION}-#{Time.now.strftime('%F-%H%M%S')}.csv"
```
maximecb, line 12: The problem with this is that it will create a large number of output files, one for every run. IMO specifying the output path as an input was the right way to go.

maximecb:
Generally not opposed to these changes, but the `run_benchmarks.rb` script would need to be updated as well, and I prefer specifying the `OUT_CSV_PATH` as an input. It's simpler and more straightforward.

eregon:
I adapted `run_benchmarks.rb`, should be good now.

eregon:
Thanks for the reviews and merging!

PR #24, opened 2021-06-23
When running Railsbench, do the setup automatically.

noahgibbs:
We care about this because if we're going to have CI-like systems for benchmarks, we need setup for individual benchmarks to happen automatically.

I also considered a little setup.sh script. But then we'd need scripts like run_benchmarks.rb to go through and run them, which I don't think is better.

### DIFF
--- a/benchmarks/railsbench/benchmark.rb
+++ b/benchmarks/railsbench/benchmark.rb
@@ -1,5 +1,17 @@
 require 'harness'
 
+# Before we activate Bundler, make sure gems are installed.
+# And before we load ActiveRecord, let's make sure the
+# database exists and is up to date.
+# Note: db:migrate will create the DB if it doesn't exist,
+# and this app's db/seeds.rb will delete and repopulate
+# the database, so rows shouldn't accumulate.
+Dir.chdir(__dir__) do
+  unless system({ 'RAILS_ENV' => 'production' }, "bundle install && bin/rails db:migrate db:seed")
+    raise "Couldn't set up railsbench!"
+  end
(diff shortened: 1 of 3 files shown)


benchmarks/railsbench/benchmark.rb
```diff
@@ -1,5 +1,17 @@
 require 'harness'
 
+# Before we activate Bundler, make sure gems are installed.
+# And before we load ActiveRecord, let's make sure the
+# database exists and is up to date.
+# Note: db:migrate will create the DB if it doesn't exist,
+# and this app's db/seeds.rb will delete and repopulate
+# the database, so rows shouldn't accumulate.
+Dir.chdir(__dir__) do
+  unless system({ 'RAILS_ENV' => 'production' }, "bundle install && bin/rails db:migrate db:seed")
```
maximecb, line 10: Is the `db:create` command from above unnecessary?

noahgibbs: In recent Rails versions if you run "bin/rails db:migrate" it will automatically create the database, yes. That wasn't true in earlier versions of Rails. I specifically tested with our railsbench.

I *am* having a bit of trouble with the "bundle install" half of this, though, on my AWS instance. So I'm going to mark this as draft and ask you not to merge it until I can track that part down.

noahgibbs:
This now works on both my Mac and the AWS instance *if* a PR to update RailsBench for recent prerelease Ruby is also applied. (https://github.com/Shopify/yjit-bench/pull/25)

PR #25, opened 2021-06-24
Add email-related gems needed by Rails that Ruby is removing

noahgibbs:
It appears I tested incorrectly - not only does the latest Rails not fix the gem problem we're seeing, but in fact Rails cannot directly add those as a dependency because of a weird double-inclusion problem.

So for now, we'll need to add the three email-related gems used by Rails that used to be builtins (net-smtp, net-pop, net-imap) to our Gemfile. That's also what Rails is doing for newly-created Rails apps in latest head-of-master.

### DIFF
--- a/benchmarks/railsbench/Gemfile
+++ b/benchmarks/railsbench/Gemfile
@@ -31,6 +31,11 @@ gem 'jbuilder', '~> 2.7'
 
 gem 'psych', '~> 3.3.2'
 
+# For Ruby 3.1.0 and higher, we need to include these gems ourselves:
+gem 'net-smtp', '~> 0.2.1'
+gem 'net-pop', '~> 0.1.1'
+gem 'net-imap', '~> 0.2.1'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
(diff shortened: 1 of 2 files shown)


noahgibbs:
I should downgrade Bundler. I haven't been careful with Bundler version in the Gemfile, and I don't think the newer Bundler should be needed to add these gems.

Here's a link to the double-inclusion problem: https://bugs.ruby-lang.org/issues/17873

Short version (as I understand it) is that the not-built-into-Ruby versions of these gems depend on other not-built-into-Ruby versions of gems, which conflict with the versions built into Ruby :-( The digest gem is a dependency here that causes this problem, though I can't recall if there are others.

You can see Rails choosing to use the same fix (adding it to each individual app's Gemfile, not gems-as-deps-of-Rails-gem) here: https://github.com/rails/rails/blob/acee501edd0274acb1b073aac1ac41a3181aafdd/Gemfile#L172-L182

eregon:
Thanks for the information. This seems all unfortunate.

The error from above seems to come from `io-wait` 0.1.1 not existing on RubyGems: https://rubygems.org/gems/io-wait
And probably it only exists as a default gem of Ruby 3.1-dev.

Issue #26, opened 2021-07-06
Bundler version in Gemfile.lock should be compatible with earlier Rubies


noahgibbs:
I used the Bundler built into YJIT (Ruby 3.1 prerelease) when updating the Gemfile.lock last time, and Bundler gets picky about that. It's a much better idea retain compatibility with other Rubies.

It would be *nice* to just allow any Bundler version, but that requires either hand-editing Gemfile.lock on every commit or using a very old Bundler version exclusively when updating it. Both "fixes" are quite annoying. On the plus side, Gemfile.lock is really simple and hand-editable.

Anyway: for now I'll pick an older Bundler version and specify it in Gemfile.lock.

noahgibbs:
It looks like Bundler is just printing a warning for this, so I'm going to ignore it. I'm not finding substantial incompatibilities *because* of this, just *near* this :-)

PR #27, opened 2021-07-07
Bench setup fixes

noahgibbs:
Several small fixes here:

Make sure Gemfile.lock files have a "Bundled with" of an actual released Bundler, in this case 2.1.4. See issue #26.

My auto-setup fix in railsbench was causing self-suspends (the process sends itself a SIGSTOP) in some cases because bash interactive shells do that when something asks for console input. I've filed an issue with the Dev team for a long-term fix (https://github.com/Shopify/dev/issues/6286), but hard to say if they'll consider it important. It's weird to me that I can find more complaints about it, but so far I can't.

For railsbench we should check in schema.rb, like normal for a Rails app. This also removes an annoying "new file" in git status.

For the jekyll test, it was failing in some cases (duplicate "Mark" top-level class) because it picked up different gem versions on different runs -- it was using an embedded Gemfile, and no Gemfile.lock. So I've checked in a Gemfile.lock from an AWS instance where it's working fine.

### DIFF
--- a/benchmarks/railsbench/benchmark.rb
+++ b/benchmarks/railsbench/benchmark.rb
@@ -7,10 +7,16 @@ require 'harness'
 # and this app's db/seeds.rb will delete and repopulate
 # the database, so rows shouldn't accumulate.
 Dir.chdir(__dir__) do
-  # Use the user's current shell and bash -i to make sure this works in Shopify Mac dev tools.
+  chruby_stanza = ""
+  if ENV['RUBY_ROOT']
+    ruby_name = ENV['RUBY_ROOT'].split("/")[-1]
+    chruby_stanza = "chruby #{ruby_name} && "
+  end
+
+  # Source Shopify-located chruby if it exists to make sure this works in Shopify Mac dev tools.
   # Use bash -l to propagate non-Shopify-style chruby config.
-  # Note that this snippet can run in bash, fish or zsh, so expand it only with great care.
-  unless system({ 'RAILS_ENV' => 'production' }, "#{ENV['SHELL']} -il -c 'bundle install && bin/rails db:migrate db:seed'")
+  success = system({ 'RAILS_ENV' => 'production' }, "/bin/bash -l -c '[ -f /opt/dev/dev.sh ] && . /opt/dev/dev.sh ]; #{chruby_stanza}bundle install && bin/rails db:migrate db:seed'")
+  unless success
     raise "Couldn't set up railsbench!"
   end
 end
--- a/benchmarks/jekyll/test-three-zero/.gitignore
+++ b/benchmarks/jekyll/test-three-zero/.gitignore
@@ -9,7 +9,6 @@
 # Ruby Gem
 *.gem
 .bundle
-Gemfile.lock
 **/vendor/bundle
 
 # Node.js and NPM
--- a/benchmarks/jekyll/test-three-zero/Gemfile.lock
+++ b/benchmarks/jekyll/test-three-zero/Gemfile.lock
@@ -1,0 +1,87 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    addressable (2.8.0)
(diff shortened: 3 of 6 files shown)


noahgibbs:
Huh. I'm not seeing anything obviously using Git protocol in any Gemfile.lock in yjit-bench. There *is* a package.json that uses a Git-protocol source (benchmarks/jekyll/test-three-zero/package.json uses git://github.com/mmistakes/minimal-mistakes.git). I'll see if it's easy to fix the package.json.

I'll also investigate if caution-tape-bot might be flagging something I didn't think to check.

noahgibbs:
Nope, this caution-tape-bot rule definitely doesn't trigger on package.json, only Gemfile.lock. I'm not sure how it's detecting only git: urls, though -- it looks like it's triggering on any added URL. https://github.com/Shopify/services-db/blob/main/components/caution_tape_bot/app/models/caution_tape_bot/rules/restrict_gem_protocol.rb

benchmarks/railsbench/benchmark.rb
```diff
@@ -7,10 +7,16 @@
...
+  if ENV['RUBY_ROOT']
+    ruby_name = ENV['RUBY_ROOT'].split("/")[-1]
+    chruby_stanza = "chruby #{ruby_name} && "
+  end
+
+  # Source Shopify-located chruby if it exists to make sure this works in Shopify Mac dev tools.
   # Use bash -l to propagate non-Shopify-style chruby config.
-  # Note that this snippet can run in bash, fish or zsh, so expand it only with great care.
-  unless system({ 'RAILS_ENV' => 'production' }, "#{ENV['SHELL']} -il -c 'bundle install && bin/rails db:migrate db:seed'")
+  success = system({ 'RAILS_ENV' => 'production' }, "/bin/bash -l -c '[ -f /opt/dev/dev.sh ] && . /opt/dev/dev.sh ]; #{chruby_stanza}bundle install && bin/rails db:migrate db:seed'")
```
maximecb, line 18: Does this rely on dev being installed on people's machines?

maximecb:
I've been finding caution tape bot really unhelpful because it comes and posts comments in issues but it never actually links you to where the supposed issue would be in the code. I don't think you really need to worry about this too much because the benchmarking code is not security or mission critical.

benchmarks/railsbench/benchmark.rb
noahgibbs: Nope. It sources the file if it exists and does nothing if it doesn't.

noahgibbs: Specifically: the dash-f in brackets followed by double ampersand means "do a file-existence test on the file, and only do the thing following if it's true." Since I haven't set fail-on-error for this script I *could* have just sourced the file and it would still work okay, but it's more polite not to leave an error message for non-Dev-users, even if this would still work.

noahgibbs:
That's fair. In this case I was able to locate the rule in question and it looks like it's a bug in that rule. It'll actually complain about *any* Git protocol, including https. So this wasn't a case of me overlooking a real problem (I worry about those.)

noahgibbs:
Ah, that rule is very recently added to Caution Tape Bot. I've filed an issue (https://github.com/Shopify/services-db/issues/8554) to get it fixed.

noahgibbs:
Converting to draft, as I realised there's at least one important setup task that isn't working and I hadn't noticed -- the Lee benchmark fails if you don't have the Victor gem installed, but its Gemfile.lock doesn't work on Ruby 3.1+, so we can't just bundle install. Working on a fix now.

noahgibbs:
Okay, fixed that. Moving forward.

ryanwilsonperkin:
Apologies for the caution tape bot noise, thank you for creating the issue. Looks like the new rule is too broad and shouldn't be matching on this

maximecb:
Thank you Noah 🙏

PR #33, opened 2021-08-04
stackprof is only available on CRuby currently

eregon:
With that small change, railsbench just works on truffleruby-dev (from `ruby-build truffleruby-dev ~/.rubies/truffleruby-dev`), with e.g.:
```
cd benchmarks/railsbench
WARMUP_ITRS=100 MIN_BENCH_ITRS=50 MIN_BENCH_TIME=0 ruby -I../../harness benchmark.rb
```
cc @noahgibbs


noahgibbs:
Looks good.

Issue #34, opened 2021-08-05
The OptCarrot benchmark used in this repo leaks Fibers


eregon:
I noticed the OptCarrot benchmark used in this repo does something quite unusual:
https://github.com/Shopify/yjit-bench/blob/6a2dd4a5ad3b653e9ce3c9d169b3b11d33b09e28/benchmarks/optcarrot/benchmark.rb#L4-L8

That creates a new `Optcarrot::NES` every time, and that also means it leaks one Fiber per instance/run.
CRuby has the feature/behavior/bug to hard kill (skip `ensure` code) Fibers when they are no longer reachable (not sure how reliable it is though), but that's in general not really feasible without native coroutines: https://github.com/oracle/truffleruby/issues/959#issuecomment-357229190
And it's a proper mess semantically-speaking.

I would suggest to either:
* Run OptCarrot as it meant to be (AFAIK), i.e., try to achieve >= 60 FPS, so reuse the same Optcarrot::NES instance and keep rendering more frames. The OptCarrot [README](https://github.com/mame/optcarrot/#benchmark-example) shows a 3000 frames plot to show peak performance. Lan_Master.nes has basically a consistent frames loop after some startup.
* Fix OptCarrot to actually dispose the [PPU Fiber](https://github.com/mame/optcarrot/blob/15def556c305d457d9b50f0a6ec9460fa1695ca1/lib/optcarrot/ppu.rb#L878) so the benchmark is more sensible to run with multiple Optcarrot::NES instances. That still means some extra IO etc (to load the cartridge) and maybe some other issues as the code seems not designed to be run like that.

maximecb:
I wouldn't want to just run more frames because that's going to make the behavior different for every iteration of the benchmark. If there's an easy fix to kill the fibers we could merge that.

eregon:
I see, that's a fair concern.

I'll see what's possible to release the PPU fiber, or to properly reset the NES.
There is NES#reset, and if that is reliable it would be one way to create a single NEW instance and avoid loading the cartridge every time, and avoid creating many Fibers during the benchmark.
Right now that doesn't reset `@frame` but that's trivial to add. I'll check if that works.

I was curious if the frames would follow some repeating pattern and actually I don't see any obvious repetition.
However, it's always the same 4 frames after startup (7-10), and there is only 10 different screens displayed (for 3000 frames):
https://gist.github.com/eregon/f80614e4435ff779585f3e65ac1bfa61

eregon:
[...]

PR #35, opened 2021-08-07
Update to mame/optcarrot@54c2098

eregon:
* Fixes https://github.com/Shopify/yjit-bench/issues/34
* By removing the directory, cloning upstream, and keeping `benchmark.rb` and `run_many.rb`.


noahgibbs (approved):

benchmarks/optcarrot/tools/plot.rb
```diff
@@ -9,58 +9,116 @@
...
-      :
-        "Start-up time (the time to show the first frame)"
+
+    d = mean + std
+    break_start = max = d.max.max.to_f.ceil(-1) + 10
+    if frames == 3000
+      d = mean + std
+      break_start = d[d.index != "truffleruby"].max.max.to_f.ceil(-1) + 10
+      d = mean - std
+      break_end = d[d.index == "truffleruby"].min.min.to_f.floor(-1) - 10
```
maximecb, line 39: What does this do?

noahgibbs: Looks like this is its commit of origin in optcarrot: https://github.com/mame/optcarrot/commit/765d9ed6b4916fe43dfd44d994ed253a33e220a3

Not 100% sure, but I know some of their graphs break out TruffleRuby separately since its scale messes with the axis height. This looks like it may be what does that.

noahgibbs: I'm guessing this is the output graph that the Truffle-specific break is mean to be part of:

https://github.com/mame/optcarrot/blob/master/doc/benchmark-summary-3000.png

eregon: Yes. I imported the latest Optcarrot master.
The actual change regarding Fibers is https://github.com/mame/optcarrot/pull/35

eregon:
Let's merge this?

noahgibbs:
She's back now, so we should get that soon.

PR #36, opened 2021-08-09
Railsbench logging improvements

noahgibbs:
* Reduce Railsbench default log level so we're not benchmarking so much logging code
* Remove prod logs at the start of each benchmark run to avoid filling up disks

With :info as the logging setting, the runs I did this weekend (15 runs, 250 iters/run) were generating a log of over 5 GB before filling the disk and needing to be restarted. Not only is that inconvenient for long runs, it also means we're burning time on logging. That isn't mostly what we care about benchmarking.


benchmarks/railsbench/benchmark.rb
```diff
@@ -19,6 +19,9 @@
   unless success
     raise "Couldn't set up railsbench!"
   end
+
+  # Over time, it's possible to fill a device with logfiles
+  system("rm -f #{__dir__}/log/production.log")
```
eregon, line 24: Wouldn't `File.unlink path if File.exist? path` be cleaner here?
Notably `system` doesn't raise anything if the process fails.

noahgibbs: Could mean some kind of filesystem error, as you say. Could mean we don't have permissions set up properly. Could mean something's wrong in the shell. Mostly removing the log is a usually-optional action meant for unusual circumstances, so ignoring failure seems reasonable.

noahgibbs: It's actually quite common to have a logfile on the system. There's normally nothing wrong with that. This is being removed only because we don't want to accumulate logfiles over a lot of runs since that can eventually fill up the disk. That's not going to be as common with the log level turned down -- :info got very large.

maximecb:
@casperisfine could you tell us which logging level is typically used for rails in production?

noahgibbs:
Do we actually log at :info in production? The other places I've worked have generally reduced that, often to :warning or :error. :info produces quite large logs.

casperisfine:
> Do we actually log at :info in production

Yes we do. Without `INFO` you wouldn't even get the requests parameters etc.

maximecb:
So you would suggest: log at info, but set log path to /dev/null for the benchmark.

casperisfine:
> log at info, but set log path to /dev/null for the benchmark.

Indeed. I think it would be the most representative in term of actual CPU work, while also getting rid of the disk access which contributes to randomness in performance.

casperisfine:
Also the actual logging backend varies a tons from companies to companies. At Shopify we have some complex kafka based stuff, some companies use SysLog, other good old log files, Heroku logs to STDOUT, etc etc.

maximecb (approved):

casperisfine (approved):

PR #40, opened 2021-09-06
Add benchmark of mail gem parsing

jhawthorn:
This adds a benchmark of parsing an email with the [`mail` gem](https://github.com/mikel/mail/). I think this is a good way to test generated parsers from [ragel](http://www.colm.net/open-source/ragel/) as some of the ones in mail [are huge](https://github.com/mikel/mail/blob/master/lib/mail/parsers/address_lists_parser.rb) 😅.

### DIFF
--- a/benchmarks/mail/benchmark.rb
+++ b/benchmarks/mail/benchmark.rb
@@ -1,0 +1,17 @@
+require "bundler/inline"
+require "harness"
+
+gemfile do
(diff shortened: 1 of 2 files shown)


noahgibbs (approved):

noahgibbs:
Looks good to me. Very simple, which is not a bad thing.

PR #42, opened 2021-09-13
Add Shopify-required files for an open-source project

noahgibbs:
Shopify has some internal requirements for all OSS projects (https://development.shopify.io/engineering/overview/approach/open_source#2_0_Starting_a_new_open_source_project_with_Shopify_code). As I'm getting yjit-metrics set up, seems like a good time to add them for yjit-bench too.

The license file and code of conduct are both pretty much boilerplate (I'm assuming we want MIT license.) The CONTRIBUTING file is a little more freeform and I'm very open to changes in phrasing. But this seems like a decent start.

And apparently we were supposed to turn on a bot that collects CLAs?

### DIFF
--- a/.github/probots.yml
+++ b/.github/probots.yml
@@ -1,0 +1,2 @@
+enabled:
+  - cla
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,80 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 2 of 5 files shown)


maximecb:
> And apparently we were supposed to turn on a bot that collects CLAs?

I tried that in the YJIT repo and it miserably failed when trying to merge commits from upstream, because it wanted all the upstream committers to agree to CLAs. After discussing this with Mike, we decided to leave it off. I think this is an unrealistic requirement.

PR #43, opened 2021-09-14
Update nbody to mame's version

jhawthorn:
Slightly faster version via https://twitter.com/mametter/status/1437595050565718022

I'm not sure if we want to update this or not, but as we currently show no improvement on nbody and only jit ~7% it seems like a good time.

### DIFF
--- a/benchmarks/nbody/benchmark.rb
+++ b/benchmarks/nbody/benchmark.rb
@@ -1,9 +1,9 @@
 # The Computer Language Benchmarks Game
 # https://salsa.debian.org/benchmarksgame-team/benchmarksgame/
 #
-# Optimized for Ruby by Jesse Millikan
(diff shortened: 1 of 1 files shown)


maximecb:
The code is not very different. Definitely doesn't seem like it would be more efficient.

Is this the version typically used to benchmark CRuby and other Ruby JITs?

jhawthorn:
It's new but I assume it would be the version being looked at since it's the fastest on https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/nbody-ruby-3.html

I was mostly interested in this from the tweet

maximecb:
Okay ;)

maximecb:
Looks like with your changes we get 99.9% coverage on it

PR #45, opened 2021-09-28
Benchmark for keyword argument sends

kddnewton:


maximecb:
@noahgibbs we're adding this tiny new benchmark to assess the kwarg optimizations we're about to make.

PR #51, opened 2021-10-18
Move Dir.chdir so that bundler/setup picks up the correct Gemfile

noahgibbs:


eregon:
Good catch, I didn't run into that as I tend to run benchmarks from their directory and not the git repo root.
It seems a good idea in general to first `cd`/`Dir.chdir` in that benchmark's dir for a script running multiple benchmarks.

PR #56, opened 2021-11-04
Add Gemfile setup to yjit-bench, including all existing harnesses

noahgibbs:
Since we want all harnesses to do this in basically the same way, I've added a harness/harness-common.rb file to add use_gemfile.

I tried having use_gemfile grovel through the caller stack and set the directory. It worked. It meant one less line of boilerplate in the calling benchmarks. But I don't love having a method named use_gemfile set the directory as a side effect, and we *do* want to set the directory for each benchmark. The benchmarks that don't set their directory have often had bugs when run from an unexpected directory. So we *should* be setting the directory in basically every benchmark -- I just don't think use_gemfile is the right place to do that. And we should do it in (at least) every benchmark that touches a file or requires gems.

yjit-metrics does *not* load the same harness-common.rb, partly because it specifies the Bundler version when initialising Gemfiles. So it has a separate implementation of use_gemfile.

This is tested with all benchmarks in yjit-bench and with all harnesses, though not with every benchmark *in* every harness. I've also tested them with the new yjit-metrics harness locally. With luck this will fix up benchmark CI as well, as it removes bundler/inline and some of the Bundler version difficulties.

The version of minitest specified by Lee's Gemfile is finicky about which Ruby version. But we *do* want to install victor when running Lee. So I decided the most reasonable compromise was to remove minitest, a dev-only gem, from its Gemfile here in yjit-bench.

### DIFF
--- a/benchmarks/activerecord/Gemfile.lock
+++ b/benchmarks/activerecord/Gemfile.lock
@@ -1,0 +1,34 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    activemodel (6.0.4.1)
+      activesupport (= 6.0.4.1)
+    activerecord (6.0.4.1)
+      activemodel (= 6.0.4.1)
+      activesupport (= 6.0.4.1)
+    activesupport (6.0.4.1)
+      concurrent-ruby (~> 1.0, >= 1.0.2)
+      i18n (>= 0.7, < 2)
+      minitest (~> 5.1)
+      tzinfo (~> 1.1)
+      zeitwerk (~> 2.2, >= 2.2.2)
+    concurrent-ruby (1.1.9)
+    i18n (1.8.11)
+      concurrent-ruby (~> 1.0)
+    minitest (5.14.4)
+    sqlite3 (1.4.2)
+    thread_safe (0.3.6)
+    tzinfo (1.2.9)
+      thread_safe (~> 0.1)
+    zeitwerk (2.5.1)
+
+PLATFORMS
+  x86_64-darwin-20
+
+DEPENDENCIES
+  activerecord (~> 6.0.3, >= 6.0.3.6)
+  activerecord-jdbcsqlite3-adapter (~> 60.4)
+  sqlite3 (~> 1.4)
+
+BUNDLED WITH
+   2.2.22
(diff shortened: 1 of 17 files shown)


eregon (approved):
This looks great.
Agreed it's best if every benchmark explicitly `Dir.chdir` for clarity.

harness/harness-common.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  chruby_stanza = ""
+  if ENV['RUBY_ROOT']
+    ruby_name = ENV['RUBY_ROOT'].split("/")[-1]
+    chruby_stanza = "chruby && chruby #{ruby_name} && "
+  end
+
+  # Source Shopify-located chruby if it exists to make sure this works in Shopify Mac dev tools.
+  # Use bash -l to propagate non-Shopify-style chruby config.
+  cmd = "/bin/bash -l -c '[ -f /opt/dev/dev.sh ] && . /opt/dev/dev.sh; #{chruby_stanza}bundle install'"
```
eregon, line 13: I think we'll be able to do something much simpler for non-Shopify-dev cases here after this PR, I'll try that once this is merged.

benchmarks/activerecord/Gemfile.lock
```diff
@@ -0,0 +1,34 @@
...
+PLATFORMS
+  x86_64-darwin-20
+
+DEPENDENCIES
+  activerecord (~> 6.0.3, >= 6.0.3.6)
+  activerecord-jdbcsqlite3-adapter (~> 60.4)
+  sqlite3 (~> 1.4)
+
+BUNDLED WITH
+   2.2.22
```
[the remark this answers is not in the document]
noahgibbs: Right now I'm assuming the Gemfile.lock bundler versions are going to be kind of a mess. In yjit-metrics I now specify a Bundler version (2.2.30) that should work fine back to 2.7 and 2.6, but I'm also careful not to assume the Gemfile.lock matches.

harness/harness-common.rb
```diff
@@ -0,0 +1,25 @@
...
+  if extra_bundled_setup
+    cmd += " && #{extra_bundled_setup}"
+  end
```
maximecb, line 16: Wondering if `extra_setup_cmd` might be a more descriptive variable name?

maximecb: Also, if `use_gemfile` does more than just taking care of a gemfile, could also consider renaming `use_gemfile` to `benchmark_setup` or something along those lines? What do you think?

maximecb (commented):
Looks good. Left two minor comments but this seems like a nice improvement to the benchmark setup code 👍

maximecb (approved):

PR #60, opened 2021-11-18
Erubi and erubi_rails benchmarks

noahgibbs:
I've added two Erubi benchmarks. The one called "erubi" uses an Erb template from the "gem server" command with real gemspec data from my local Ruby install.

The erubi_rails benchmark uses an Erb template from Discourse's topics/show view, stubbed with fake data. It's rendered via a tiny Rails test app.

### DIFF
--- a/benchmarks/erubi/benchmark.rb
+++ b/benchmarks/erubi/benchmark.rb
@@ -2,19 +2,24 @@ require 'harness'
 Dir.chdir __dir__
 use_gemfile
 
+# The current version is index.erb for the "gem server" command.
+# I got it from https://github.com/jnewland/sinatra-rubygems, a
+# reimplementation in Sinatra.
 TEMPLATE_FILE = "simple_template.erb"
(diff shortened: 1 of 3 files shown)


noahgibbs:
It depends on Rails, much as Railsbench does. It doesn't actually *run* Rails' HTTP handling, though - we're rendering a single template outside the normal HTTP loop. It doesn't even run a controller action, nor hook up its only controller to any routes.

It has a lot of files, of course, since it's a Rails app. But the amount of code that isn't from "rails new" is very small.

That specific Discourse template won't work well outside Rails because it uses Rails-specific template functionality like content_for. I expect I could find some smaller Discourse template that rendered correctly outside Rails.

maximecb:
Ok. We can merge it as-is, if you feel comfortable maintaining it.

noahgibbs:
Definitely. This is much easier than Railsbench :-)

PR #63, opened 2021-11-24
Use frozen_string_literal: true with the generated Erubi source code

noahgibbs:
If I add a strategic comment when defining the method for erubi, we get *much* better YJIT optimisation.

I'm a little worried about just obscuring this case. But if we want to fix it by changing our benchmark, this does it.

### DIFF
--- a/benchmarks/erubi/benchmark.rb
+++ b/benchmarks/erubi/benchmark.rb
@@ -44,7 +44,7 @@ template = File.read TEMPLATE_FILE
 source = generate_source(template)
 
 # Create a method with the generated source
-eval "def run_erb; #{source}; end"
+eval "# frozen_string_literal: true\ndef run_erb; #{source}; end"
 
 # This is taken from actual "gem server" data
 @values = JSON.load(File.read "gem_specs.json")


flavorjones:
We should upstream that to Erubi!

maximecb:
Is that the way rails does it though? Do they use frozen_string_literal when evaling erb template code?

noahgibbs:
Mike: Erubi will cheerfully generate code with that label, but it doesn't work if you put it inside a method - the frozen_string_literal comment must be the first in the file or eval.

noahgibbs:
I'm looking through action_view (https://github.com/rails/rails/tree/main/actionview) to see if Rails does this. It looks like anything with a layout does this already (https://github.com/rails/rails/blob/main/actionview/lib/action_view/layouts.rb#L323). I can't tell if "bare" Erb with no templates does, though. They're not passing the freeze argument when constructing an Erubi Engine (https://github.com/rails/rails/blob/main/actionview/lib/action_view/template/handlers/erb.rb and https://github.com/rails/rails/blob/main/actionview/lib/action_view/template/handlers/erb/erubi.rb). But mostly they seem to be adding their own annotations to source in a lot of places, so that's not strongly indicative either way.

Doing some searching, though, I'm *not* seeing anything else that would obviously put in frozen_string_literal: true as the first line. It looks like the file is read directly to provide the "source" parameter to an UnboundTemplate, to a Template, and only encoding is changed during those steps. There's a "preamble" passed to Erubi, but it's used for tagging output with which template file it came from in debug mode, not adding a frozen_string_literal.

So I *think* that means Rails isn't passing frozen_string_literal when doing its module_eval to define the template method (https://github.com/rails/rails/blob/main/actionview/lib/action_view/template.rb#L300).

noahgibbs:
Since that's what Rails does, that means *our* Rails benchmark probably is *not* passing that comment. But it's still being optimised reasonably. I'm not 100% sure why that is.

noahgibbs:
At this point I don't think there's anything to merge. Closing. If anybody has strong opinions they can reopen or file a new issue.

PR #65, opened 2021-11-29
Use RbConfig.ruby for Ruby

jhawthorn:
This avoids depending on chruby or dev.

This works on my machine 😅 but I'm not sure if it will cause problems for others (especially those using Shopify's "dev")

### DIFF
--- a/harness/harness-common.rb
+++ b/harness/harness-common.rb
@@ -1,21 +1,15 @@
+def run_cmd(*args)
+  puts "Command: #{args.join(" ")}"
+  system(*args)
+end
+
 # Set up a Gemfile, install gems and do extra setup
 def use_gemfile(extra_setup_cmd: nil)
   # Benchmarks should normally set their current directory and then call this method.
 
-  chruby_stanza = ""
-  if ENV['RUBY_ROOT']
-    ruby_name = ENV['RUBY_ROOT'].split("/")[-1]
-    chruby_stanza = "chruby && chruby #{ruby_name} && "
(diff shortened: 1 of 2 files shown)


eregon:
I think it could be even simpler than that.
It seems fair to assume that the ruby being used is the Ruby in PATH, right?

Then maybe we should just check
```ruby
abort unless File.realpath(RbConfig.ruby) == File.realpath(`which ruby`.chomp)
```

And simply `system "bundle", ...`, etc.

noahgibbs:
I know Maxime had worried about yjit-metrics and this PR. Good news on that count: yjit-metrics implements its own harness with the same API, so changes here that don't change the API won't break yjit-metrics.

I'll check this and get back to you Monday about how the code here interacts with Dev and zsh. Though I don't expect any significant zsh/bash differences here.

noahgibbs:
Cool. This is working fine for me on Mac with dev, trying a few different benchmarks and Rubies. I'll look at adding the same changes to the yjit-metrics harness, but that shouldn't slow down merging this.

I'm having trouble with erubi_rails and YJIT because it fails installing puma... But I don't think that's about this change. I don't see the same problem with Ruby 3.0.2 if I uninstall Puma first. So the gem path stuff is probably fine.

noahgibbs (approved):

noahgibbs:
I've gone ahead and merged this as-is, and I'll add the RUBY_DESCRIPTION change shortly.

PR #66, opened 2021-11-30
Make the harness work with Fish

tenderlove:
Previously we assumed chruby was available in bash, but some people (me)
install it on Fish.

This commit just adds support for Fish shell.  I was pretty conservative about shell detection, and there may be a better way.  But this gets the benchmarks running on my Linux machine and my work Mac using Fish.  I'm definitely open to ways to make this better!


tenderlove:
I just realized this probably breaks on zsh? 😅

eregon:
IMHO the way to go is to not mess with any shell here, and just use subcommands directly:
https://github.com/Shopify/yjit-bench/pull/65#issuecomment-981939751

tenderlove:
Ya, I agree, #65 is a better solution.  Closing in favor of that PR

Issue #68, opened 2022-01-02
The erubi_rails benchmark defines new classes in every request


eregon:
This initialization code, https://github.com/Shopify/yjit-bench/blob/9d56ed48fbfdb04904657d9dde8249e4561ec0b0/benchmarks/erubi_rails/app/controllers/fake_discourse_controller.rb#L18-L44, is run once for every request (verified by adding `p [:FakeDiscourseController_init, object_id, self]` as the first line of `initialize`).
That is likely unintended, and as a result there are 10 `Struct.new` per request, which involves creating new classes, defining methods dynamically, etc.

Here is the output I got on 3.1.0:
```
$ cd benchmarks/erubi_rails
$ ruby -I../../harness $PWD/benchmark.rb
ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
[:FakeDiscourseController_init, 15640, #<FakeDiscourseController:0x00000000007a30>]
[:FakeDiscourseController_init, 15660, #<FakeDiscourseController:0x00000000007a58>]
[:FakeDiscourseController_init, 15780, #<FakeDiscourseController:0x00000000007b48>]
[:FakeDiscourseController_init, 15820, #<FakeDiscourseController:0x00000000007b98>]
[:FakeDiscourseController_init, 15860, #<FakeDiscourseController:0x00000000007be8>]
[:FakeDiscourseController_init, 15900, #<FakeDiscourseController:0x00000000007c38>]
[:FakeDiscourseController_init, 15940, #<FakeDiscourseController:0x00000000007c88>]
[:FakeDiscourseController_init, 15980, #<FakeDiscourseController:0x00000000007cd8>]
[:FakeDiscourseController_init, 16020, #<FakeDiscourseController:0x00000000007d28>]
[:FakeDiscourseController_init, 16060, #<FakeDiscourseController:0x00000000007d78>]
[:FakeDiscourseController_init, 16100, #<FakeDiscourseController:0x00000000007dc8>]
[:FakeDiscourseController_init, 16140, #<FakeDiscourseController:0x00000000007e18>]
[:FakeDiscourseController_init, 16180, #<FakeDiscourseController:0x00000000007e68>]
[...]

PR #72, opened 2022-01-04
Cache erubi_rails stubs between requests

noahgibbs:
This is intended to address issue #68, pointed out by @eregon.

### DIFF
--- a/benchmarks/erubi_rails/app/controllers/fake_discourse_controller.rb
+++ b/benchmarks/erubi_rails/app/controllers/fake_discourse_controller.rb
@@ -15,18 +15,19 @@ SiteSetting = ss_struct.new(false, company_struct.new(false), "settings title")
 class FakeDiscourseController < ApplicationController
     attr_reader :stub_assigns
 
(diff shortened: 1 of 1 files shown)


eregon:
Looks good, I'll try it soon.

FWIW I was thinking a separate class (or even just top-level) might make sense, and e.g., just store the result in a constant or so or pass it in, given there is `fake_controller.stub_assigns` in https://github.com/Shopify/yjit-bench/blob/1751916ceb5de60311b6c632caed34f68d2a7982/benchmarks/erubi_rails/benchmark.rb#L17
Maybe the `@stub_assigns` is used too though?
Anyway this looks fine.

noahgibbs:
Yeah, I don't have a strong opinion there. I think putting it into the controller file makes sense rather than adding a separate file for it, but within that file I don't have a strong opinion where it should be.

noahgibbs:
Sure. I was waiting a bit to see if anybody else wanted to. But this seems pretty safe.

PR #75, opened 2022-01-26
In railsbench and erubi_rails, always force "ruby" platform

noahgibbs:
In railsbench and erubi_rails, specify that we should always use platform ruby. This will guarantee we compile all C extensions from scratch, no prebuilt.

This is because for prerelease Rubies, prebuilt C extensions can have a wrong Ruby C ABI and that won't be detected, as a rule. This *can* happen for any benchmark that uses a gem with native extensions with a prerelease Ruby version. It definitely happens in railsbench and erubi_rails.

The fix in this PR is not fully adequate - whenever we use a prerelease Ruby, it turns out we need to never use anything prebuilt. Ruby's ABI can change without detection, and probably has since the gem prebuilt binary was built. @flavorjones is considering whether it makes sense to propose a Bundler change to do this automatically for dev/prerelease Rubies. @peterzhu2118's proposed ABI change also seems relevant here to detect (but not fix) the problem.

For benchmark CI I'm going to configure our user to always use the Ruby platform (no prebuilt) for *everything*. It already always fully reinstalls all gems, Rubygems, etc on every build. That approach, plus never using prebuilt, should fix all benchmarks at a basically-tolerable speed penalty. Shouldn't affect timing for benchmark iterations, just non-graphed overhead for the job.

### DIFF
--- a/benchmarks/railsbench/.bundle/config
+++ b/benchmarks/railsbench/.bundle/config
@@ -1,0 +1,2 @@
+---
+BUNDLE_FORCE_RUBY_PLATFORM: "true"
--- a/benchmarks/railsbench/.bundle/config
+++ b/benchmarks/railsbench/.bundle/config
@@ -1,2 +1,4 @@
 ---
+# To work correctly with native extensions and prerelease Ruby, we'll
+# want to always build all native extensions from source, never prebuilt.
 BUNDLE_FORCE_RUBY_PLATFORM: "true"
--- a/benchmarks/erubi_rails/.bundle/config
+++ b/benchmarks/erubi_rails/.bundle/config
@@ -1,0 +1,2 @@
+---
+BUNDLE_FORCE_RUBY_PLATFORM: "true"
--- a/benchmarks/erubi_rails/.bundle/config
+++ b/benchmarks/erubi_rails/.bundle/config
@@ -1,2 +1,4 @@
 ---
+# To work correctly with native extensions and prerelease Ruby, we'll
+# want to always build all native extensions from source, never prebuilt.
 BUNDLE_FORCE_RUBY_PLATFORM: "true"


benchmarks/railsbench/.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_FORCE_RUBY_PLATFORM: "true"
```
maximecb, line 2: And the same comment here

eregon:
I don't quite understand this change, there are no prebuilt gems for ruby-head anyway, isn't it?

Would it make more sense to do this as part of the harness-common setup, maybe only when the current Ruby is a prerelease?
It'd end up creating this file though, so I don't think it would change much, except it would apply to all benchmarks automatically.

Another possibility, which would only affect ruby-head would be to set `BUNDLE_FORCE_RUBY_PLATFORM=true` in ENV before bundle commands, that could then be done only for ruby-head and no other cases.

Note that @peterzhu2118's PR is giving a proper ABI version for ruby-head, that should be enough for Bundler to use separate dirs for different builds, *if* `bundle config --local path` is used (Bundler already include ABI as part of the path it uses).
If `bundle config --local path` is not used and GEM_HOME is not set, then gems would end under the built's ruby prefix, which then would get cleaned up if you `rm -rf ruby-head-prefix` between builds.
If GEM_HOME is set (incorrectly, uses RUBY_VERSION and not the ABI ...) by chruby 0.3.9, then you'd want `rm -rf ruby-head-prefix` + `rm -rf ~/.gem/ruby/3.2.0` between builds.

In any case, I don't think building from source helps here (and already the case anyway), if a gem is already installed and not cleaned up between builds of different ruby-head revisions it'll be reused because already installed. The used gem home needs to be removed between builds (whether it's under vendor/bundle, in the ruby prefix or ~/.gem/ruby/3.2.0).

flavorjones:
I want to take this opportunity to clarify a few things.

> arguably that's a Nokogiri bug

Nothing in here is a Nokogiri bug. Nokogiri's `ruby` platform gem installs fine against ruby-head:

```
$ ruby -v
ruby 3.2.0dev (2022-01-18T14:00:58Z flavorjones-mkmf-p.. 58944d8b98) [x86_64-linux]

$ gem install nokogiri
Fetching nokogiri-1.13.1.gem
Building native extensions. This could take a while...
Successfully installed nokogiri-1.13.1
1 gem installed
```

And Nokogiri's native gems **correctly** set the `required_gem_version` range to exclude any version of Ruby which was not precompiled. Unfortunately, Bundler's behavior is such that for this Gemfile:

``` ruby
# Gemfile
source "https://rubygems.org"
gem "loofah"
```

if you first `bundle install` or `bundle lock` on a version of Ruby supported by the native gem (e.g, 3.1.0, where this works fine) and *then* using the existing Gemfile.lock try to `bundle install` using a version of Ruby that is **not** supported by the native gem, you'll see this error:

``` text
$ bundle install
Fetching gem metadata from https://rubygems.org/.......
Your bundle is locked to nokogiri (1.13.1-x86_64-linux) from rubygems repository https://rubygems.org/ or installed
locally, but that version can no longer be found in that source. That means the author of nokogiri
(1.13.1-x86_64-linux) has removed it. You'll need to update your bundle to a version other than nokogiri
(1.13.1-x86_64-linux) that hasn't been removed in order to install.
```

<details><summary>Here's a whole user journey demonstrating the issue</summary>
# Attempt 1

``` ruby
# Gemfile
source "https://rubygems.org"
gem "loofah"
```

## ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]

``` text
$ bundle install
Fetching gem metadata from https://rubygems.org/........
Resolving dependencies...
Using bundler 2.3.6
Using racc 1.6.0
Using crass 1.0.6
Fetching nokogiri 1.13.1 (x86_64-linux)
Installing nokogiri 1.13.1 (x86_64-linux)
Using loofah 2.13.0
Bundle complete! 1 Gemfile dependency, 5 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
```

``` text
# Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    crass (1.0.6)
    loofah (2.13.0)
      crass (~> 1.0.2)
      nokogiri (>= 1.5.9)
    nokogiri (1.13.1-x86_64-linux)
      racc (~> 1.4)
    racc (1.6.0)

PLATFORMS
  x86_64-linux

DEPENDENCIES
  loofah

BUNDLED WITH
   2.3.6
```

## ruby 3.2.0dev (2022-01-18T14:00:58Z flavorjones-mkmf-p.. 58944d8b98) [x86_64-linux]

``` text
$ bundle install
Fetching gem metadata from https://rubygems.org/.
Bundler 2.4.0.dev is running, but your lockfile was generated with 2.3.6. Installing Bundler 2.3.6 and restarting using that version.
Fetching bundler 2.3.6
Installing bundler 2.3.6
Fetching gem metadata from https://rubygems.org/.......
Your bundle is locked to nokogiri (1.13.1-x86_64-linux) from rubygems repository https://rubygems.org/ or installed
locally, but that version can no longer be found in that source. That means the author of nokogiri
(1.13.1-x86_64-linux) has removed it. You'll need to update your bundle to a version other than nokogiri
(1.13.1-x86_64-linux) that hasn't been removed in order to install.
```

# Attempt 2

## ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]

``` text
$ bundle lock --add-platform=ruby
```

``` text
# Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    crass (1.0.6)
    loofah (2.13.0)
      crass (~> 1.0.2)
      nokogiri (>= 1.5.9)
    mini_portile2 (2.7.1)
    nokogiri (1.13.1)
      mini_portile2 (~> 2.7.0)
      racc (~> 1.4)
    nokogiri (1.13.1-x86_64-linux)
      racc (~> 1.4)
    racc (1.6.0)

PLATFORMS
  ruby
  x86_64-linux

DEPENDENCIES
  loofah

BUNDLED WITH
   2.3.6
```

``` text
$ bundle install
Using bundler 2.3.6
Using crass 1.0.6
Using racc 1.6.0
Using nokogiri 1.13.1 (x86_64-linux)
Using loofah 2.13.0
Bundle complete! 1 Gemfile dependency, 5 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
```

## ruby 3.2.0dev (2022-01-18T14:00:58Z flavorjones-mkmf-p.. 58944d8b98) [x86_64-linux]

``` text
$ bundle install
Fetching gem metadata from https://rubygems.org/.......
Your bundle is locked to nokogiri (1.13.1-x86_64-linux) from rubygems repository https://rubygems.org/ or installed
locally, but that version can no longer be found in that source. That means the author of nokogiri
(1.13.1-x86_64-linux) has removed it. You'll need to update your bundle to a version other than nokogiri
(1.13.1-x86_64-linux) that hasn't been removed in order to install.
```
</details>


@deivid-rodriguez has acknowledged in the past (I can't find the issue right now) that Bundler does not support this use case -- sharing Gemfile.lock solutions across multiple versions of Ruby. One solution is to blow away the Gemfile.lock when changing Ruby versions. Another is to force `ruby` platform gems at all time, which is what this PR does.

deivid-rodriguez:
Yes, I would like to support that properly but it's not currently fully supported. Sharing a lockfile across ruby versions does no behave well if the ruby version requirements for a locked gem change between those versions. From what a gathered last time I looked, the problem is that ruby version requirements are not currently recorded in the `Gemfile.lock` file like other dependencies. I opened this ticket when I first found this: https://github.com/rubygems/rubygems/issues/4455.

I will have a look at this and see if I can make it work better.

eregon:
Thanks a lot for the explanation, that's a rather unfortunate situation.
I guess the older "resolve to binary gem every time" approach of Bundler worked better for this case, rather than specific binaries/platforms ending up in Gemfile.lock.
And agreed multiple Gemfiles per benchmark is total overkill.
I think this workaround makes sense then.
What might also work would be to create the Gemfile.lock with ruby-head and not a release, but it's more fragile as one might recreate the Gemfile.lock with a release accidentally.

I suggest to update the comment for clarity to something like:
```
# Force to install from source for prerelease Ruby, otherwise the Gemfile.lock will choose
# a binary gem for e.g. nokogiri (even though that's only supported for releases), which fails at bundle install time
```

noahgibbs:
At this point we're handling this by just carefully inspecting Gemfiles as they're added. I think we can close this.

PR #82, opened 2022-03-07
Add chunky_png benchmark based on main one for gem

noahgibbs:
Looks like YJIT gives just over 30% speedup here - 680-ish ms with YJIT versus 1010-ish without.

Chunky_png was the interesting category of benchmark in the old Bench9000 benchmark that I didn't feel like we already had. But that benchmark code was GPL'd, so I started from the MIT-licensed code in the chunky_png gem instead.

### DIFF
--- a/benchmarks/chunky_png/Gemfile
+++ b/benchmarks/chunky_png/Gemfile
@@ -1,0 +1,3 @@
+source "https://rubygems.org"
+
+gem "chunky_png"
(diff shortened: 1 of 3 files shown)


maximecb:
Neat. I like that our benchmark suite is pretty diverse and this seems like a nice addition.

I'm assuming this calls our to a C library and is not implementing PNG compression in Ruby right? Would be cool if it did hehe.

noahgibbs:
My understanding is that chunky_png is all Ruby. It has no non-Ruby dependencies: https://github.com/wvanbergen/chunky_png

maximecb:
Oh that's cool. Lots of potential for massive speedups in the future.

PR #88, opened 2022-04-27
Add a str_concat microbenchmark

noahgibbs:
This is very small, and just for checking optimisation of the << operator on strings.

Since changing the capacity had no significant effect on speed, I added a comment to that effect and removed the capacity testing.

### DIFF
--- a/benchmarks/str_concat.rb
+++ b/benchmarks/str_concat.rb
@@ -1,26 +1,32 @@
+# frozen_string_literal: true
 require 'harness'
 
 # Structuring this with a single method means we will not
 # get a single-encoding call site. That doesn't matter now,
 # but may matter with more-optimised generated code.
 def concat_single_test(n, encoding, str_to_add)
-    # We used to supply capacity when building a string, but so far
-    # it makes only a very small difference - around 5.08 vs 5.09 sec
-    # for 10k iterations. Maybe add it back when/if we've optimised
-    # significantly more?
-    s = String.new(encoding: encoding)
-    n.times { s << str_to_add }
+  # We used to supply capacity when building a string, but so far
+  # it makes only a very small difference - around 5.08 vs 5.09 sec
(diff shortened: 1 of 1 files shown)


noahgibbs:
Dropping to 100 iterations (because I don't want to cut/paste 10k str concats). First off, I get significant run-to-run variation.

No YJIT:

Times: 68k iters/sec
While loop: 79k iters/sec
cut-and-paste inlined: 88k iters/sec

With YJIT:

Times: 69k iters/sec
While loop: 87k iters/sec
cut-and-paste inlined: 87k iters/sec

So YJIT seems to be optimising the while loop to basically as fast as the inlining, though non-JIT CRuby runs it slightly slower.

I'm seeing a speed difference on some runs, but it's (e.g.) 87,365 iters/sec vs 88,682 iters/sec. Definitely not worth it for the added complexity of manually inlining, at least with YJIT.

The while loop is a significant speedup over times and easy to add, so I'll do that.

noahgibbs:
Also, with a benchmark this micro, a surprising number of things matter. For instance, assigning the string to a constant and re-concatenating from the constant every time actually speeds it up very noticeably. Oy.

maximecb:
Let's go with the while loop and no manual inlining as you suggested.

Speaking of constants, this reminds me we should be setting `# frozen_string_literal: true`.

maximecb (approved):

maximecb:
They must have changed the concat algorithm.

I'd be curious to work out how many microseconds and clock cycles per concatenation it takes. For example if we concatenate 10K times, and the benchmark takes 10ms to run (10,000 us), that would be 1us per concatenation. Assuming a 3GHz clock cycle, that would be

`(1/1,000,000) * 3 * 10^9 = 3000 clock cycles`

maximecb:
That's actually not bad, though I'm sure we can make it faster :)

PR #92, opened 2022-05-24
Run DB tasks with the same RAILS_ENV

k0kubun:
I think `bin/rails db:migrate db:seed` needs to be run with `RAILS_ENV=production` too. It returns 500 because of this.


noahgibbs:
Ah, we've had trouble with this one. Thank you!

(LGTM)

maximecb:
Much appreciated!

PR #107, opened 2022-08-15
Update README to point to ruby/ruby YJIT build instructions

jemmaissroff:
The readme was describing cloning https://github.com/shopify/yjit, which no longer has yjit code. Instead, link to the build instructions on ruby/ruby.


maximecb:
Oh what the hell lol, how did we not catch this!? Thank you Jemma 🙏

PR #109, opened 2022-08-19
Make Ruby indentation consistent

k0kubun:
This repository sometimes uses 2 spaces and sometimes 4 spaces. When I asked Maxime about the same thing in bootstraptest, she said she wants to be consistent with existing CRuby code, 2 spaces.

This PR has only indentation changes. Once this PR is approved and merged, I'll add `.git-blame-ignore-revs` entry and add the commit to the file.


noahgibbs:
GitHub verifies that this is only whitespace changes: https://github.com/Shopify/yjit-bench/pull/109/files?diff=unified&w=1

noahgibbs (approved):

eregon:
The indentation of nbody is also inconsistent, with some mix of 1-space and 2-spaces: https://github.com/Shopify/yjit-bench/blob/main/benchmarks/nbody/benchmark.rb
I think it would be nice to fix it as well in this PR.

k0kubun:
> GitHub verifies that this is only whitespace changes: https://github.com/Shopify/yjit-bench/pull/109/files?diff=unified&w=1

I didn't know of this feature. Cool!

> The indentation of nbody is also inconsistent, with some mix of 1-space and 2-spaces: https://github.com/Shopify/yjit-bench/blob/main/benchmarks/nbody/benchmark.rb
> I think it would be nice to fix it as well in this PR.

https://github.com/ruby/ruby/blob/master/benchmark/so_nbody.rb has the same issue, and I was assuming we wanted to keep the original shootout code as is. But given that it says it's "modified by Yusuke Endoh", now I realized that it probably doesn't matter. I'll just fix both. https://github.com/Shopify/yjit-bench/pull/109/commits/9dc4ab19dbff88fe85c9c8e9fa7eb9a57a0ba691

PR #115, opened 2022-09-19
Run bundle install without --quiet when not satisfied

k0kubun:
`bundle install` sometimes takes a lot of time, so I'd like to monitor its progress when dependencies are unsatisfied. To keep the output quiet when satisfied, I added `bundle check ||` before `bundle install`.


kddnewton (approved):
Makes sense

noahgibbs:
LGTM.

The only ways I can think of this being worse is if Bundler is flaky or we're messing with the Gemfile while yjit-bench runs. And if those happen we have significant problems regardless :-)

PR #119, opened 2022-10-20
Add Ruby LSP benchmarks

vinistock:
### Motivation

The Ruby LSP benefits a lot from using YJIT. Further YJIT improvements can make the experience of using the Ruby LSP in the editor even better.

### Questions

There are three intensive types of operations the Ruby LSP will do
- File parsing
- Syntax Tree visitor requests
- RuboCop requests

Would it be worth separating these three into separate folders? For now, I put everything in the same benchmark block, but I'd assume that each one of those might have distinct performance bottlenecks.

Also, should I commit the `Gemfile.lock` file? I noticed some benchmarks do and others don't.

### DIFF
--- a/benchmarks/ruby-lsp/.rubocop.yml
+++ b/benchmarks/ruby-lsp/.rubocop.yml
@@ -1,0 +1,6 @@
+require:
+  - rubocop-performance
+  - rubocop-rails
+
+AllCops:
+  NewCops: disable
--- a/benchmarks/ruby-lsp/Gemfile
+++ b/benchmarks/ruby-lsp/Gemfile
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "ruby-lsp"
+gem "rubocop"
+gem "rubocop-performance"
+gem "rubocop-rails"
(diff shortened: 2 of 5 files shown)


k0kubun:
> Also, should I commit the Gemfile.lock file? I noticed some benchmarks do and others don't.

"others" seem to be optcarrot and rubykon. They don't need `bundle exec` to run the benchmark. It looks like every benchmark that depends on dependencies in Gemfile has Gemfile.lock, so I think it's good that this PR has Gemfile.lock.

k0kubun (approved):

noahgibbs:
Right now we don't have an easy way to return multiple speed measurements from the same benchmark. To have multiple, you'd need to separate them into three benchmark names (e.g. ruby-lsp-parsing, ruby-lsp-syntax, ruby-lsp-rubocop.) I don't think we want multiple LSP benchmarks right now.

+1 to Kokubun's comments about Gemfile.lock.

maximecb:
Thanks for putting this together @vinistock

It's true that different tasks may have different performance bottlenecks. We could start with this one benchmark for now though, because if we have too many, we might not take the time to look at the individual results of each.

@noahgibbs is this PR ready to merge or do we want any changes made to it?

noahgibbs:
Oh hey, looks like we're missing a file, ruby-lsp/internal. So this is breaking. @vinistock is this easy to fix?

PR #132, opened 2022-11-16
Document how to run a single benchmark

eregon:
I have found that for my usage of yjit-bench where e.g. I want to benchmark TruffleRuby or more Ruby versions, `run_benchmarks.rb` assumes way too much about the platform, which Rubies are benchmarked and more.

OTOH a huge advantage of this repository/set of benchmarks is they are all extremely easy to run with just `ruby -Iharness benchmarks/some_benchmark.rb` so I documented that. That's in fact I think much better than many other benchmarks suites, because it's so easy to run a benchmark and get started.
And then one can of course build some fairly trivial scripts to run N/all benchmarks on M Rubies etc based on their needs.


noahgibbs:
LGTM

PR #133, opened 2022-11-16
Preallocate a planet object

tenderlove:
In Ruby, we introduced a shape transition based on the capacity of the object.  Objects will track their max size on the class, and subsequent allocations will pre-allocate room.  This means that two instances of the same class could be born with different shape ids.

This change ensures that the planet objects in the list all have the same shape ids so that the ivar inline caches can be monomorphic.

More discussion is on this PR: https://github.com/ruby/ruby/pull/6699

### DIFF
--- a/benchmarks/nbody/benchmark.rb
+++ b/benchmarks/nbody/benchmark.rb
@@ -82,6 +82,10 @@ def offset_momentum(bodies)
   b.vz = - pz / SOLAR_MASS
 end
 
+# Pre-allocate a planet so that subsequent Planet objects
+# will have the same shape.
+Planet.new(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0)
+
 BODIES = [
   # sun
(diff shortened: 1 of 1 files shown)


benchmarks/nbody/benchmark.rb
```diff
@@ -82,6 +82,10 @@ def offset_momentum(bodies)
   b.vz = - pz / SOLAR_MASS
 end
 
+# Pre-allocate a planet so that subsequent Planet objects
+# will have the same shape.
+Planet.new(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0)
```
eregon, line 87: From my knowledge of maps/Shapes, this sounds like a hack and a bug of the Shape transition implementation.

There is a set of `Planet.new` calls below, all with the same arguments, all setting the same ivars.
If they don't have the same Shape, it's a bug of the Shape system.

noahgibbs:
As an alternative here, if there's some specific shape behaviour you'd like to make sure is benchmarked over time, you could submit a new benchmark.

PR #135, opened 2022-11-18
Add a benchmark that sets instance variables to an object

tenderlove:
Our other setivar benchmark sets instance variables to an immediate value.  We can kind of cheat because if the JIT compiler is able to detect that the value we write is an immediate, then it can omit any write barrier code.

This benchmark writes an object to the instance variable.  The goal of this benchmark is to measure the impact of the write barrier.  Hopefully we can skip executing the write barrier in some cases, and this benchmark will help us measure the impact.


noahgibbs:
LGTM

PR #138, opened 2022-11-22
Unset GEM_HOME and GEM_PATH for chruby

k0kubun:
https://github.com/Shopify/yjit-bench/pull/112 had been working fine when I was using rbenv. After I switched to chruby, I started to see weird errors in the version that I use for running run_benchmarks.rb. 

The difference between rbenv and chruby was that chruby sets GEM_HOME and GEM_PATH in your environment, which must be unset to make `bundle install` with https://github.com/Shopify/yjit-bench/pull/112 work.


eregon (approved):
Looks good to me.
As an aside, I wish chruby wouldn't set those env vars (not the discussion here though).

PR #139, opened 2022-11-22
Get rid of end_time

k0kubun:
https://github.com/Shopify/yjit-bench/pull/136 moved the place where `end_time` is printed, but I think he didn't intend to make that change.

I usually share yjit-bench results like this https://github.com/ruby/ruby/pull/6790, and I've intentionally skipped `end_time` so that you can focus on the RUBY_DESCRIPTION of target rubies. It's nice to show `end_time` first so that I don't need to manually remove that line after copy-pasting it.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -324,16 +324,16 @@ check_pstate()
 # Create the output directory
 FileUtils.mkdir_p(args.out_path)
 
-metadata = {}
+ruby_descriptions = {}
 args.executables.each do |name, executable|
-  metadata[name] = check_output([*executable, "-v"]).chomp
+  ruby_descriptions[name] = check_output([*executable, "-v"]).chomp
 end
 
 # Benchmark with and without YJIT
 bench_start_time = Time.now.to_f
 bench_times = {}
 args.executables.each do |name, executable|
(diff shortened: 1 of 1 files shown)


run_benchmarks.rb
```diff
@@ -318,7 +318,7 @@ def run_benchmarks(ruby:, ruby_description:, name_filters:, out_path:, pre_init:
 # Create the output directory
 FileUtils.mkdir_p(args.out_path)
 
-metadata = {}
+metadata = { end_time: Time.now.strftime("%Y-%m-%d %H:%M:%S %Z (%z)") }
```
eregon, line 321: See https://github.com/Shopify/yjit-bench/pull/136#discussion_r1027908435
This wouldn't capture the time at the right time.

But maybe this end_time should just be removed, cc @noahgibbs ?

k0kubun: ugh yeah okay. But, given that Noah (who maintains yjit-metrics) is not sure if we should maintain this, I'm in favor of just removing it. Even when I was running this script multiple times in batches, I didn't utilize that. Just showing "Total time spent benchmarking" seems enough in most cases.

k0kubun: removed it https://github.com/Shopify/yjit-bench/pull/139/commits/74e4cde1d5b1c31a88f1e023c72217299bd72d17

noahgibbs: Yeah, I'm fine with removing it.

PR #140, opened 2022-11-22
Skip db:seed if already done

k0kubun:
`created_at` and `updated_at` are part of the JSON responses, so to make sure it will be a fair comparison, I think we should use the same timestamps across different Ruby commands too.

If you agree with that idea, I'll backport this to the original railsbench as well.


noahgibbs:
That looks good. Given how slow the inserts are, I'm all for this :-)

I had thought about just checking that the data *existed*, but of course that's not correct. This is a better approach.

PR #142, opened 2022-12-02
Sort benchmarks like speed.yjit.org

k0kubun:
I noticed that "Benchmarks Speed Details" in the [Latest Full Details](https://speed.yjit.org/benchmarks/bench-2022-12-02-041023) page has a table like this harness and it shows headlining benchmarks first, other benchmarks second, and micro-benchmarks third.

In order for people to focus on more important benchmarks, I think it's nice to show results in that order in this default harness as well.

### DIFF
--- a/benchmarks.yml
+++ b/benchmarks.yml
@@ -1,0 +1,70 @@
+headline:
+  # Real-esque benchmarks that you could pretend are real for a blog post or a paper
+  activerecord:
+    desc: activerecord repeatedly queries entries in a SQLite table with highly-random names.
+  psych-load:
+    desc: psych-load repeatedly loads a small selection of YAML files taken from various OSS projects.
+  mail:
+    desc: mail tests the Mail gem by repeatedly creating an email from a text file and converting it to a string for sending.
+  liquid-render:
+    desc: liquid-render renders a chosen-for-profiling Liquid theme repeatedly.
+  jekyll:
+    desc: jekyll reviews and rebuilds a Jekyll site, but is almost entirely scanning directories of files that didn't change.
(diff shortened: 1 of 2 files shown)


k0kubun:
> I'm not thrilled to have a second place we keep a list of headlining benchmarks.
>
> Having it there always felt kind of clunky and inelegant to me. Keeping a duplicate list is in another repo is even worse.

I agree. I first thought of adding yjit-metrics as a git submodule and referring to `BENCHMARK_METADATA`, but also thought it's a circular dependency that would make it tedious to add a new benchmark. The dependency should always flow from yjit-metrics to yjit-bench, not the other way around. But it's not possible to fix it in a single PR. I'd be happy to start moving that in this PR.

> Maybe we could move the benchmark description into yjit-bench somehow, in the benchmark itself? Comments in the file, or something like that? WDYT?
>
> I was thinking a small YAML file next to benchmark.rb would work nicely
> It could also be a single YAML file at the root, or as you said a header in the benchmark file itself

Defining a constant to hold metadata in a script would unnecessarily impact the benchmark. However, writing it as a comment and extracting it feel a little bit hacky / less maintainable compared to just separately maintaining the metadata. Some benchmarks only have a single `desc` currently, so having a separate file only for that feels too much. So I vote for maintaining something similar to `BENCHMARK_METADATA` as a single YAML file in yjit-bench.

maximecb:
> I'd suggest adding a `path` value which is the benchmark file, so this YAML file also serves as an easy way to iterate all benchmark files. I think the outer grouping by category makes the format a bit harder to evolve and to process, I'd just make it an Array at the top-level and avoid the nesting. It's small things though, feel free to ignore.

Those are good ideas. It would be nice to have an array of benchmarks, with each benchmark having a `path`, and a `category` field.

PR #144, opened 2022-12-08
Add --rss option to compare RSS

k0kubun:
[speed.yjit.org](https://speed.yjit.org/) shows memory usages on "Latest Full Details". This PR makes [the same thing](https://github.com/Shopify/yjit-metrics/blob/main/metrics-harness/harness.rb#L138-L142) available on local yjit-bench.

### Interpreter: Ruby 3.1 vs 3.2
```
ruby3.1: ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]
ruby3.2: ruby 3.2.0dev (2022-12-07T23:30:00Z master a2d3f5606a) [x86_64-linux]

----------  ------------  ----------  ---------  ------------  ----------  ---------  ---------------  ---------------
bench       ruby3.1 (ms)  stddev (%)  RSS (MiB)  ruby3.2 (ms)  stddev (%)  RSS (MiB)  ruby3.1/ruby3.2  ruby3.2 1st itr
railsbench  1637.3        1.0         104.2      1566.2        0.9         103.3      1.05             1.03
----------  ------------  ----------  ---------  ------------  ----------  ---------  ---------------  ---------------
```

### YJIT: Ruby 3.1 vs 3.2
```
yjit3.1: ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) +YJIT [x86_64-linux]
yjit3.2: ruby 3.2.0dev (2022-12-07T23:30:00Z master a2d3f5606a) +YJIT [x86_64-linux]

----------  ------------  ----------  ---------  ------------  ----------  ---------  ---------------  ---------------
bench       yjit3.1 (ms)  stddev (%)  RSS (MiB)  yjit3.2 (ms)  stddev (%)  RSS (MiB)  yjit3.1/yjit3.2  yjit3.2 1st itr
railsbench  1226.1        2.1         372.8      1084.8        1.9         119.4      1.13             1.06
----------  ------------  ----------  ---------  ------------  ----------  ---------  ---------------  ---------------
```

### DIFF
--- a/harness/harness.rb
+++ b/harness/harness.rb
@@ -13,6 +13,8 @@ MIN_BENCH_TIME = ENV.fetch('MIN_BENCH_TIME', 10).to_i
 default_path = "data/results-#{RUBY_ENGINE}-#{RUBY_ENGINE_VERSION}-#{Time.now.strftime('%F-%H%M%S')}.csv"
 OUT_CSV_PATH = File.expand_path(ENV.fetch('OUT_CSV_PATH', default_path))
 
+RSS_CSV_PATH = ENV['RSS_CSV_PATH'] ? File.expand_path(ENV['RSS_CSV_PATH']) : nil
+
 system('mkdir', '-p', File.dirname(OUT_CSV_PATH))
 
 puts RUBY_DESCRIPTION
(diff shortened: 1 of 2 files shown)


casperisfine:
Might be just me having a biff with RSS, but I wonder if we could find a better memory metric in `/proc/self/smaps_rollup`.

k0kubun:
It doesn't work on macOS, right? As written in Noah's original source code comment, one benefit of this implementation is that it works on both macOS and Linux. I use both macOS and Linux for running yjit-bench locally, so I appreciate the current implementation working on both platforms.

Also I want yjit-bench to be consistent with [speed.yjit.org](https://speed.yjit.org/). If you want to see a metric that's not RSS at the end of iterations, I yjit-metrics to be changed first.

casperisfine:
Fair 👍

PR #111, opened 2022-08-20
Introduce -e option to override executables

k0kubun:
## Summary
Introduce `-e` option that works like:

```rb
# Benchmark interp and yjit, compare interp/yjit (backward compatible)
./run_benchmarks

# Benchmark interp and yjit, compare interp/yjit (same thing, being explicit for an example)
./run_benchmarks -e "interp::ruby" -e "yjit::ruby --yjit"

# Benchmark yjit only
./run_benchmarks -e "yjit::ruby --yjit"

# Benchmark yjit only (just skipped the name, which also works)
./run_benchmarks -e "ruby --yjit"

# Benchmark mjit and yjit, compare mjit/yjit
./run_benchmarks -e "mjit::ruby --mjit" -e "yjit::ruby --yjit"
```

Note that CRuby's `make benchmark` (benchmark_driver.gem) accepts `-e` option with the same interface.

## Example
It's backward-compatible:

```
$ ./run_benchmarks.rb getivar
...

end_time: 2022-08-19 17:06:37 PDT (-0700)
interp: ruby 3.2.0dev (2022-08-19T14:31:14Z yjit_backend_ir c6247485c4) [arm64-darwin21]
yjit: ruby 3.2.0dev (2022-08-19T14:31:14Z yjit_backend_ir c6247485c4) +YJIT [arm64-darwin21]

-------  -----------  ----------  ---------  ----------  ----------------  ------------
bench    interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit (ms)  yjit 1st itr
getivar  116.3        0.5         27.6       0.4         4.22              1.00
-------  -----------  ----------  ---------  ----------  ----------------  ------------
Legend:
- interp/yjit: ratio of interp/yjit time. Higher is better for yjit. Above 1 represents a speedup.
- yjit 1st itr: ratio of interp/yjit time for the first benchmarking iteration.
```

A single executable:

```
$ ./run_benchmarks.rb getivar -e "ruby --yjit"
...

end_time: 2022-08-19 17:08:20 PDT (-0700)
ruby --yjit: ruby 3.2.0dev (2022-08-19T14:31:14Z yjit_backend_ir c6247485c4) +YJIT [arm64-darwin21]

-------  ----------------  ----------
bench    ruby --yjit (ms)  stddev (%)
getivar  27.8              2.9
-------  ----------------  ----------
```

MJIT vs YJIT:

```
[...]

PR #146, opened 2022-12-09
Add --chruby option to use a Ruby managed by chruby

k0kubun:
This is a shorthand for `-e` option https://github.com/Shopify/yjit-bench/pull/111, which works like:

```bash
# same as: ./run_benchmarks.rb -e "3.1.3::/opt/rubies/3.1.3/bin/ruby"
./run_benchmarks.rb --chruby 3.1.3

# same as: ./run_benchmarks.rb -e "yjit::/opt/rubies/ruby-master/bin/ruby --yjit"
./run_benchmarks.rb --chruby "yjit::ruby-master --yjit"
```

Similar to `-e`, this is the same interface as [what ruby/ruby's benchmark driver has](https://github.com/benchmark-driver/benchmark-driver/commit/d318a353a77b68c3cb419554761b99238f194550).

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -303,6 +303,18 @@ OptionParser.new do |opts|
     args.executables[name] = executable.shellsplit
   end
 
+  opts.on("--chruby=NAME::VERSION OPTIONS", "ruby version under chruby and options to be benchmarked") do |v|
+    name, version = v.split("::", 2)
+    if version.nil?
+      version = name # allow skipping `NAME::`
+    end
+    version, *options = version.shellsplit
+    unless executable = ["/opt/rubies/#{version}/bin/ruby", "#{ENV["HOME"]}/.rubies/#{version}/bin/ruby"].find { |path| File.executable?(path) }
+      abort "Cannot find '#{version}' in /opt/rubies or ~/.rubies"
+    end
+    args.executables[name] = [executable, *options]
+  end
+
   opts.on("--out_path=OUT_PATH", "directory where to store output data files") do |v|
     args.out_path = v
   end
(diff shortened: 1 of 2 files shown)


run_benchmarks.rb
```diff
@@ -303,6 +303,18 @@ def run_benchmarks(ruby:, ruby_description:, name_filters:, out_path:, pre_init:
     args.executables[name] = executable.shellsplit
   end
 
+  opts.on("--chruby=NAME::VERSION OPTIONS", "ruby version under chruby and options to be benchmarked") do |v|
+    name, version = v.split("::", 2)
+    if version.nil?
+      version = name # allow skipping `NAME::`
+    end
+    version, *options = version.shellsplit
+    unless executable = ["/opt/rubies/#{version}/bin/ruby", "#{ENV["HOME"]}/.rubies/#{version}/bin/ruby"].find { |path| File.executable?(path) }
```
noahgibbs, line 312: Ah, this is a good fix. I had tried a similar thing with running chruby, but it's hard to get the environment right.

This won't change the gem environment, though, right? So it would only work if the Rubies are both using the same built native extensions?

eregon: It's https://github.com/k0kubun/yjit-bench/blob/9fc2cd39d2ddebdb58f0ae7965530390a225cc7f/run_benchmarks.rb#L263-L267
And that's only done if it's not the first `ruby` in PATH, I missed that.
So then the behavior seems:
* If it's the first Ruby in PATH use GEM_HOME/GEM_PATH as set by chruby or anything else
* Otherwise use the default gem home inside the prefix

PR #153, opened 2022-12-16
Test that benchmarks run correctly on CI

k0kubun:
This PR makes sure benchmark scripts in this repository work correctly when we change them on CI. We also test a couple of benchmark runners, run_benchmarks.rb and run_once.sh.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,9 +16,9 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.1'
+          ruby-version: 'head'
       - name: Run benchmarks for testing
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
+name: test
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
```
eregon, line 10: FWIW this is already the default: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

maximecb:
Could you also test the `./run_once.sh` script with one benchmark while we're at it? :)

maximecb (approved):

PR #154, opened 2022-12-16
Dockerfile for Ruby v3.2.0-rc1 and latest jyit-bench commit

echiugoog:
Dockerfile using alpine 3.16 as base image for Ruby 3.2.0-rc1 and latest yjit-bench


docker/Dockerfile
```diff
@@ -0,0 +1,187 @@
...
+	rm -rf /build && \
+# verify we have no "ruby" packages installed
+	if \
+		apk --no-network list --installed \
+			| grep -v '^[.]ruby-rundeps' \
+			| grep -i ruby \
+	; then \
+		exit 1; \
+	fi; \
+	[ "$(command -v ruby)" = '/usr/local/bin/ruby' ]
```
maximecb, line 142: I'd like some comment(s) explaining what this is doing

maximecb (commented):
I like the idea of having a dockerfile for yjit-bench, but this is my first time hearing about alpine linux.

I think I'd prefer we base it off of a mainstream distribution like Ubuntu. If only because it will make it easier for a broader range of people to be able to troubleshoot this if anything breaks.

I would also like to see more commenting for the various steps. A lot is happening in this dockerfile.

docker/Dockerfile
echiugoog: Much of this came from the original file (now added a reference and LICENSE file to comply)

What was added in was a loop to go thru Gemfiles in yjit-bench and preinstall them into the image so it's not done at runtime.

docker/Dockerfile
```diff
@@ -0,0 +1,187 @@
...
+COPY --from=builder /usr/local/ /usr/local/
+ADD patch-05-tzinfo-data /tmp/
+ADD patch-06-tzinfo-data /tmp/
+
+RUN git clone https://github.com/Shopify/yjit-bench.git /yjit-bench && \
+	cd /yjit-bench \
+	git reset --hard ${YJIT_COMMIT} && \
+	\
+	patch -p1 < /tmp/patch-05-tzinfo-data && \
+	patch -p1 < /tmp/patch-06-tzinfo-data
```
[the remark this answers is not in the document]
echiugoog: I had issues with tzinfo-data gem, seems to be resolved by installing the tzdata alpine package now

echiugoog:
Alpine is compact lightweight linux distro that's popular for building container images. I renamed the Dockerfile with an alpine suffix. Can add in a ubuntu one as well. The build should be fairly similar.

docker/LICENSE
```diff
@@ -0,0 +1,23 @@
+Copyright (C) 2014 Docker, Inc. All rights reserved.
+Copyright (C) 2022 Edwin Chiu
```
maximecb, line 2: Why the extra license? We aren't redistributing the docker source?

PR #156, opened 2022-12-19
Support benchmarking any Ruby implementations

k0kubun:
* Support benchmarking non-CRuby implementations with `run_benchmarks.rb` without `-e` / `--chruby` options
  * For YJIT developers, it still defaults to compare `interp` and `yjit` when you use CRuby and `--yjit` is available.
  * I think it's generally nice for Ruby implementation developers to share as many benchmarks as possible.
* Update README to explain that it's not mandatory to use `chruby`
  * Because it just runs `RbConfig.ruby`, you could use `rbenv` instead, for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,17 +16,6 @@ CSV file written by the benchmarking harness. The output is written to
 an output CSV file at the end, so that results can be easily viewed or
 graphed in any spreadsheet editor.
 
-## Running a single benchmark
-
-This is the easiest way to run a single benchmark.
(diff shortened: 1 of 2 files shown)


noahgibbs:
Mostly this looks fine. I feel a little funny about making the default "use the current Ruby, or if it supports YJIT then use it with and without YJIT." But it's documented as doing that, and it's easy to override it, so it's not really a problem.

PR #160, opened 2023-01-03
Try updating to Nokogiri 1.3.10

noahgibbs:
When Dependabot did this, it didn't seem to get failures in its PR, but after it was merged (and broke) the auto-build per commit *did* see the problem. Weird. Gonna see if a draft PR gets a build failure or no.


noahgibbs:
Sure.

Failing commit: https://github.com/Shopify/yjit-bench/commit/4c717dc2cfecd4c5c4b31520c93458e93f5133ab

Failing build step: https://github.com/Shopify/yjit-bench/actions/runs/3829100596/jobs/6515419985

Error message from bundler:

~~~
nokogiri-1.13.10-x86_64-linux requires ruby version < 3.2.dev, >= 2.6, which is
incompatible with the current version, ruby 3.3.0p-1
~~~

noahgibbs:
I doubt it makes a difference, but this was an upgrade from 1.3.6 to 1.3.10, so the problem could date back to any of several different micro versions.

flavorjones:
also see upstream bundler PR: https://github.com/rubygems/rubygems/pull/6225

PR #162, opened 2023-01-06
Use `Random` instead of `SecureRandom` for determinism.

maximecb:
`SecureRandom` uses crypto-safe randomness sources which can't be seeded, whereas we seed the `Random` gem's RNG in `harness-common.rb `This should help the ActiveRecord benchmark behave more deterministically across runs.

Following comment on #159

### DIFF
--- a/benchmarks/activerecord/benchmark.rb
+++ b/benchmarks/activerecord/benchmark.rb
@@ -1,5 +1,4 @@
 require "harness"
-require "securerandom"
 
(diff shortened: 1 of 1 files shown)


k0kubun:
Sounds better, but you still get different `created_at`/`updated_at` values every time. I wonder if we can just reuse the whole schema and records like railsbench does.

k0kubun (approved):
However, I think we should just merge this first 🙂

PR #158, opened 2022-12-21
Add harness-cstats to collect C method stats

k0kubun:
This PR adds a harness to do the same thing as https://github.com/Shopify/yjit-bench/pull/123, collecting C method statistics, whenever we want. 

### railsbench

```
$ chruby ruby; WARMUP_ITRS=0 MIN_BENCH_ITRS=1 MIN_BENCH_TIME=0 ruby -Iharness-cstats benchmarks/railsbench/benchmark.rb
ruby 3.2.0dev (2022-12-21T15:34:21Z master 398aaed2f0) [arm64-darwin22]
Command: bundle check 2> /dev/null || bundle install
The Gemfile's dependencies are satisfied
Command: bin/rails db:migrate db:seed
Using 100 posts in the database
itr #1: 11438ms
Top C loop method iterations:
  204123 Array#each
  155759 Hash#each_pair
   80714 Hash#each
   55875 Array#map
   37448 Enumerable#map
   21190 Enumerable#each_with_object
   19301 Hash#each_key
    7800 Array#reverse_each
    6031 Hash#each_value
    4064 Array#map!
    3305 Enumerable#each_with_index
    2685 Enumerator#each
    2002 Enumerable#flat_map
     170 Enumerable#reverse_each
       0 Enumerable#each_entry

Top 100 C method calls:
 1105900 Hash#[]
  453811 Hash#[]=
  350840 Hash#fetch
  348210 Module#===
  344733 String#to_s
  305403 Class#new
  215855 BasicObject#!
  197383 Kernel#is_a?
  143338 Kernel#initialize_dup
  143332 Kernel#dup
  130418 Array#each
  128396 Kernel#nil?
  127168 String#getbyte
  127116 String#==
  118219 Array#[]
  116544 #<Class:Thread>#current
  114213 Kernel#respond_to?
  106518 Hash#key?
  105752 Integer#+
  101810 Integer#<
   98206 String#empty?
   94699 Kernel#block_given?
   93628 ActiveSupport::OrderedOptions#[]
   92093 Thread#[]
   90291 Array#empty?
   82735 BasicObject#==
   75565 String#gsub!
   74273 Symbol#to_s
   71723 String#initialize
   71216 Hash#initialize_copy
   70454 Array#hash
   70357 Hash#delete
   63584 String#setbyte
   63584 Integer#^
   62658 ActionController::Metal#_request
   60265 String#<<
   58030 ActiveModel::LazyAttributeHash#delegate_hash
   56920 Kernel#hash
[...]

PR #163, opened 2023-01-06
Count Ruby method calls as well in stats harness

k0kubun:
This PR adds Ruby stats to the harness created in https://github.com/Shopify/yjit-bench/pull/158. 

It seems useful to know about Ruby call stats as well when we discuss method inlining implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ You can find several test harnesses in this repository:
 * harness-perf - a simplified harness that runs for exactly the hinted number of iterations
 * harness-bips - a harness that measures iterations/second until stable
 * harness-continuous - a harness that adjusts the batch sizes of iterations to run in stable iteration size batches
-* harness-cstats - count C method calls and C loop iterations
+* harness-stats - count method calls and loop iterations
 
 To use it, run a benchmark script directly, specifying a harness directory with `-I`:
 
--- a/harness-stats/harness.rb
+++ b/harness-stats/harness.rb
@@ -7,6 +7,7 @@ self.singleton_class.prepend Module.new {
     frames = []
     c_calls = Hash.new { 0 }
     c_loops = Hash.new { 0 }
+    rb_calls = Hash.new { 0 }
 
     method_trace = TracePoint.new(:call, :c_call, :return, :c_return) do |tp|
       # Keep track of call frames to get the caller of :b_call
@@ -18,9 +19,12 @@ self.singleton_class.prepend Module.new {
         frames.pop
       end
 
-      # Count C method calls
-      if tp.event == :c_call
+      # Count method calls
+      case tp.event
+      when :c_call
         c_calls[method_name] += 1
+      when :call
+        rb_calls[method_name] += 1
       end
     end
 
@@ -55,6 +59,15 @@ self.singleton_class.prepend Module.new {
     puts "Top #{c_calls.size} C method calls (#{'%.1f' % c_calls_ratio}% of all #{c_calls_total} calls):"
     c_calls.sort_by(&:last).reverse.first(100).each do |method, count|
       puts '%8d (%4.1f%%) %s' % [count, 100.0 * count / c_calls_total, method]
+    end
+    puts
+
+    rb_calls_total = rb_calls.sum(&:last)
+    rb_calls = rb_calls.sort_by { |_method, count| -count }.first(100)
+    rb_calls_ratio = 100.0 * rb_calls.sum(&:last) / rb_calls_total
+    puts "Top #{rb_calls.size} Ruby method calls (#{'%.1f' % rb_calls_ratio}% of all #{rb_calls_total} calls):"
+    rb_calls.sort_by(&:last).reverse.first(100).each do |method, count|
+      puts '%8d (%4.1f%%) %s' % [count, 100.0 * count / rb_calls_total, method]
     end
   end
 }


maximecb:
I like it and I think this is very useful 👍  One thing that I find a bit confusing though:

```
Top 10 block calls by C methods (100.0% of all 52000 calls):
   22000 (42.3%) Array#each
   10000 (19.2%) Hash#each_key
    6000 (11.5%) Array#map
   ...
Top 100 C method calls (93.0% of all 856021 calls):
   87000 (10.2%) Hash#[]
```

Here I'm not sure what the block calls count mean. Based on the name it sounds like the number of times a block has been called. However, the numbers seem much too low. That is, it seems improbable that a block inside `Array#each` has only been called 22K times when `Hash#[]` has been called 87K times.

So I'm assuming that `Array#each` is not counting calls to blocks (and so the number of loop iterations performed by `Array#each`), but rather counting the number of times that we call into `Array#each` as a whole, which is not as useful of a statistic. It would be more useful if we count count the number of loop iterations for each of these... Or maybe count both the number of calls to `Array#each` and also the total number of iterations (calls to a block) done by `Array#each`, and also display the average iteration count per `Array#each` call (which I assume is probably ~5 to 20 or something).

k0kubun:
> That is, it seems improbable that a block inside Array#each has only been called 22K times when Hash#[] has been called 87K times.

The numbers seem to come from `activerecord` benchmark. I wouldn't be too surprised even if an ActiveRecord benchmark turned out to be actually calling `Hash#[]` 4x more often than `Array#each` blocks. It'd be interesting to investigate why we end up with these numbers for understanding our workloads and double-checking that the trace implementation is working.

> So I'm assuming that Array#each is not counting calls to blocks

If you have this code:

```rb
total = 0
run_benchmark do
  [1, 2, 3, 4, 5].each do |i|
    total += i
  end
end
puts "total: #{total}"
```

you get:

```
$ ruby trace.rb
Top 1 block calls by C methods (100.0% of all 5 calls):
       5 (100.0%) Array#each

Top 2 C method calls (100.0% of all 6 calls):
       5 (83.3%) Integer#+
       1 (16.7%) Array#each

Top 2 Ruby method calls (100.0% of all 3 calls):
       2 (66.7%) TracePoint#disable
       1 (33.3%) TracePoint#enable
total: 15
```

and I think it's working as I intended; it's counting the call of `Array#each` itself under "C method calls" and counting the call of blocks under "block calls by C methods".

<details>
<summary>Full script used for testing</summary>

```rb
def run_benchmark
  frames = []
  c_calls = Hash.new { 0 }
  c_loops = Hash.new { 0 }
  rb_calls = Hash.new { 0 }

  method_trace = TracePoint.new(:call, :c_call, :return, :c_return) do |tp|
    # Keep track of call frames to get the caller of :b_call
    case tp.event
    when :call, :c_call
      method_name = "#{tp.defined_class}##{tp.method_id}"
      frames.push([tp.event, method_name])
    when :return, :c_return
      frames.pop
    end

    # Count method calls
    case tp.event
    when :c_call
      c_calls[method_name] += 1
    when :call
      rb_calls[method_name] += 1
    end
  end

  block_trace = TracePoint.new(:b_call) do |tp|
    caller_event, caller_method = frames.last

    # Count block calls only when the caller is a C method
    if caller_event == :c_call
      c_loops[caller_method] += 1
    end
  end

  method_trace.enable
  block_trace.enable
  yield
ensure
  block_trace.disable
  method_trace.disable

  c_loops_total = c_loops.sum(&:last)
  c_loops = c_loops.sort_by { |_method, count| -count }.first(100)
  c_loops_ratio = 100.0 * c_loops.sum(&:last) / c_loops_total
  puts "Top #{c_loops.size} block calls by C methods (#{'%.1f' % c_loops_ratio}% of all #{c_loops_total} calls):"
  c_loops.each do |method, count|
    puts '%8d (%4.1f%%) %s' % [count, 100.0 * count / c_loops_total, method]
  end
  puts

  c_calls_total = c_calls.sum(&:last)
  c_calls = c_calls.sort_by { |_method, count| -count }.first(100)
  c_calls_ratio = 100.0 * c_calls.sum(&:last) / c_calls_total
  puts "Top #{c_calls.size} C method calls (#{'%.1f' % c_calls_ratio}% of all #{c_calls_total} calls):"
  c_calls.sort_by(&:last).reverse.first(100).each do |method, count|
    puts '%8d (%4.1f%%) %s' % [count, 100.0 * count / c_calls_total, method]
  end
  puts

  rb_calls_total = rb_calls.sum(&:last)
  rb_calls = rb_calls.sort_by { |_method, count| -count }.first(100)
  rb_calls_ratio = 100.0 * rb_calls.sum(&:last) / rb_calls_total
  puts "Top #{rb_calls.size} Ruby method calls (#{'%.1f' % rb_calls_ratio}% of all #{rb_calls_total} calls):"
  rb_calls.sort_by(&:last).reverse.first(100).each do |method, count|
    puts '%8d (%4.1f%%) %s' % [count, 100.0 * count / rb_calls_total, method]
  end
end

total = 0
run_benchmark do
  [1, 2, 3, 4, 5].each do |i|
    total += i
  end
end
puts "total: #{total}"
```

</details>

maximecb:
Then my intuition is wrong and there are way more C calls and Ruby calls than block calls.

I will merge this, but I think it could also be useful to add a breakdown of C calls, Ruby calls and block calls with the percentage for each, just to guide our optimization efforts. At the moment it seems like we might be spending a lot less time doing loop iterations than we thought?

PR #165, opened 2023-01-08
Update AR benchmark to use Random, reduce seeds, and better use case

hmistry:
Updated the ActiveRecord benchmark to be consistent with the Sequel benchmark based on feedback to use Random instead SecureRandom, reduce db seeds to speed up benchmark time, and have a better real world use case upon fetching records from db.


eregon:
That benchmark already uses Random (see https://github.com/Shopify/yjit-bench/blob/main/benchmarks/activerecord/benchmark.rb), could you rebase for a simpler diff?

PR #167, opened 2023-01-11
Avoid creating wasted string objects

k0kubun:
I felt somewhat uncomfortable seeing the code of https://github.com/Shopify/yjit-bench/pull/159 and https://github.com/Shopify/yjit-bench/pull/165 since the result is pushed to the VM stack but then popped without returning it from the block. 

It seems like the intention of https://github.com/Shopify/yjit-bench/pull/165 is to "have a better real world use case". If you imagine something like a template engine with that change, the results should be concatenated to a single string buffer, not writing part of them to a wasted object.

### DIFF
--- a/benchmarks/activerecord/benchmark.rb
+++ b/benchmarks/activerecord/benchmark.rb
@@ -36,7 +36,7 @@ Post.where(id: 1).first.title
 run_benchmark(10) do
   1.upto(1000) do |i|
     post = Post.where(id: i).first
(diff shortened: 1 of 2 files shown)


noahgibbs:
So yeah, LGTM.

eregon (commented):
LGTM as well.
One thing is the string might still be escape analyzed (when JITing upto + block), if we want to avoid that we could assign to a local variable declared above `run_benchmark(10) do`.

PR #170, opened 2023-01-18
Initial add liquid-c

jimmyhmiller:
I got this working on my machine. Not sure if we have a way to do setup on each benchmark automatically or not. You need to compile the c extension before it works. 

Here's the current stat on my machine

```
interp: ruby 3.3.0dev (2023-01-17T13:01:19Z upstream/master df6b72b8ff) [arm64-darwin22]
yjit: ruby 3.3.0dev (2023-01-17T13:01:19Z upstream/master df6b72b8ff) +YJIT dev [arm64-darwin22]

--------  -----------  ----------  ---------  ----------  -----------  ------------
bench     interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit  yjit 1st itr
liquid-c  70.5         2.2         48.8       5.6         1.45         0.24        
--------  -----------  ----------  ---------  ----------  -----------  ------------
Legend:
- interp/yjit: ratio of interp/yjit time. Higher is better for yjit. Above 1 represents a speedup.
- yjit 1st itr: ratio of interp/yjit time for the first benchmarking iteration.
```


benchmarks.yml
```diff
@@ -49,6 +49,8 @@ optcarrot:
   desc: optcarrot is a functional headless NES emulator, run on a specific game cartridge for a specific number of frames.
 rubykon:
   desc: Ruby solver for Go (the boardgame.) Runs 1,000 iterations forward from an initial starting board.
+liquid-c:
+  desc: liquid-c renders a chosen-for-profiling Liquid theme repeatedly.
```
maximecb, line 53: I think we should put it in the headline category since it's used on SFR

maximecb:
If we install liquid from the gem instead of copying the liquid-c files into the repo, does that automatically compile the C extension?

@noahgibbs any input/advice?

eregon:
Yes, if the gem is part of a Gemfile like in other benchmarks it will just compile as needed and not need to vendor any files from liquid-c.

What's surprising is the liquid-render benchmark seems to vendor files currently: https://github.com/Shopify/yjit-bench/tree/main/benchmarks/liquid-render/lib
Maybe to have a fixed liquid version? But that could be ensured by just using `gem 'liquid', 'version'`.

The [performance dir](https://github.com/Shopify/yjit-bench/tree/main/benchmarks/liquid-render/performance) is needed for the benchmark but the rest shouldn't be.

maximecb:
> Yes, if the gem is part of a Gemfile like in other benchmarks it will just compile as needed and not need to vendor any files from liquid-c.

We should use a gemfile for this benchmark then, and fix the version to what is currently the latest.

jimmyhmiller:
Yeah, I asked in slack about the vendoring. I was told there to vendor. That's the only reason I did it. Pushed the non-vendored

benchmarks.yml
jimmyhmiller: Done

maximecb:
> Yeah, I asked in slack about the vendoring. I was told there to vendor. That's the only reason I did it. Pushed the non-vendored

Sorry Jimmy!

maximecb (approved):

eregon:
https://github.com/Shopify/yjit-bench/pull/175 to also not vendor files for the liquid-render benchmark, and use the same version of liquid

PR #175, opened 2023-01-21
Remove vendored liquid and make liquid and liquid-c benchmarks consistent

eregon:
* Remove liquid-render scripts redundant with benchmark.rb
* Cleanup liquid-c benchmark
* Updates liquid from 4.0.3 to 5.4.0


maximecb:
LGTM but I would like @noahgibbs to review as well.

noahgibbs:
This looks like a significant improvement. Thanks!

PR #179, opened 2023-02-23
Update the sequel benchmark to run on Ruby < 3.1.

nirvdrum:
The `Random::Formatter` functionality used by the Sequel benchmark was introduced in Ruby 3.1, which was released 14 months ago. It can be helpful to track performance over time, but currently the full benchmark suite can't be run on Ruby 3.0 and older. This also impacts JRuby and TruffleRuby releases that shipped with the Ruby 3.0 standard library.

Since we're only using `Random::Formatter` to randomly generate strings for staging data, we can replace it with our own function and allow the benchmark to run on older Ruby releases as well.

### DIFF
--- a/benchmarks/sequel/benchmark.rb
+++ b/benchmarks/sequel/benchmark.rb
@@ -1,5 +1,5 @@
 require "harness"
-require 'random/formatter'
+require "securerandom" # Provides `Random::Formatter` in Ruby 2.6+
 
 Dir.chdir __dir__
 use_gemfile


eregon:
I think before Ruby 3.1 these methods already exist, but they are defined by the `securerandom` stdlib:
```
$ ruby -v -rsecurerandom -e 'p Random.alphanumeric 30'
ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-linux]
"EiIA9dGZYt3bouBV9JhVgZnp05sqAi"
```

eregon:
BTW `Random.alphanumeric` is also used in https://github.com/Shopify/yjit-bench/blob/main/benchmarks/activerecord/benchmark.rb

benchmarks/sequel/benchmark.rb
```diff
@@ -45,11 +44,18 @@
 class Post < Sequel::Model 
 end
 
+ALPHA_NUM_POOL = ("a".."z").to_a + ("A".."Z").to_a + ("0".."9").to_a
+def random_alphanumeric(size)
+  ret = ""
+  size.times { ret << ALPHA_NUM_POOL.sample }
+  ret
+end
```
maximecb, line 52: This code should go in `harness/harness-common.rb`

benchmarks/sequel/benchmark.rb
```diff
@@ -1,5 +1,5 @@
 require "harness"
-require 'random/formatter'
+require "securerandom" # Provides `Random::Formatter` in Ruby 2.6+
```
maximecb, line 2: Just want to make sure. We want to make sure that execution is deterministic as much as possible. Does securerandom need to be seeded separately from the default Ruby RNG?

maximecb: Just need to make sure that the random alphanumeric values are the same from one execution to the next, and if not, check if there is a way to fix a seed in `harness/harness-common.rb`

maximecb: Another option would be to define our own `Random.alphanumeric` in `harness-common.rb`, which we could then use to yank the `random/formatter` and `securerandom` dependencies from this benchmark and also the activerecord benchmark.

PR #186, opened 2023-03-03
Allow multiple executables in a single option

k0kubun:
With this PR, `--chruby "ruby" --chruby "yjit::ruby --yjit"` can be written as `--chruby "ruby;yjit::ruby --yjit"` as a shorthand.

Both `-e` and `--chruby` options are what's stolen from [benchmark-driver](https://github.com/benchmark-driver/benchmark-driver), which have had the same syntax except for this. I didn't implement this feature at first since the syntax seems a bit weird at a glance.

However, using `--chruby` with yjit-bench for a while, I've needed this very often. I'd like to use this shorthand.


noahgibbs:
Looks useful to me.

PR #187, opened 2023-03-04
Add --graph option to render a graph

k0kubun:
This PR adds an option to render a graph of benchmark results using [gruff.gem](https://github.com/topfunky/gruff). While we rely on [speed.yjit.org](https://speed.yjit.org/) for historical data, I think it's sometimes nice to have a graph image in pull requests or tickets.

## Example
Note: Using `WARMUP_ITRS=4 MIN_BENCH_ITRS=2 MIN_BENCH_TIME=0` to save my time, so some results didn't have enough warmup 🙂 

### Headline
output_218

### Other
output_219

### Micro
output_220

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,8 @@ This is the only harness that uses `run_benchmark`'s argument, `num_itrs_hint`.
 
 `--graph` option of `run_benchmarks.rb` allows you to render benchmark results as a graph.
 
-```
+```bash
+# Write a graph at data/output_XXX.png (it will print the path)
 ./run_benchmarks.rb --graph
 ```
 
@@ -210,6 +211,17 @@ brew install imagemagick
 
 # Ubuntu
 sudo apt-get install libmagickwand-dev
+```
+
+### Changing font size
+
+You can regenerate a graph with `misc/graph.rb`, changing its font size.
+
+```
+Usage: misc/graph.rb [options] CSV_PATH
+        --title SIZE                 title font size
+        --legend SIZE                legend font size
+        --marker SIZE                marker font size
 ```
 
 ## Disabling CPU Frequency Scaling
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -192,6 +192,26 @@ This is the only harness that uses `run_benchmark`'s argument, `num_itrs_hint`.
 ./run_benchmarks.rb --rss
 ```
 
+## Rendering a graph
+
+`--graph` option of `run_benchmarks.rb` allows you to render benchmark results as a graph.
+
+```
+./run_benchmarks.rb --graph
```
maximecb, line 200: Should document where people should expect to find the generated image file.

README.md
```diff
@@ -192,6 +192,26 @@ This is the only harness that uses `run_benchmark`'s argument, `num_itrs_hint`.
...
+
+Before using this option, you might need to install the dependencies of [Gruff](https://github.com/topfunky/gruff):
+
+```bash
+# macOS
+brew install imagemagick
+
+# Ubuntu
+sudo apt-get install libmagickwand-dev
+```
```
maximecb, line 213: I don't like adding extra dependencies, and from experience, imagemagick can be a pain to install on some platforms. Will allow it because it's useful.

k0kubun:
>  If the font size was a little bigger

It depends on which font. Making marker font size bigger seemed sometimes problematic.

Anyway, for that use case, I went the extra mile and carved out a standalone command to adjust the font size https://github.com/Shopify/yjit-bench/pull/187/commits/cf535036b263350f4b55e58d2270570ed5c25e86.

README.md
k0kubun: addressed in https://github.com/Shopify/yjit-bench/commit/cf535036b263350f4b55e58d2270570ed5c25e86

PR #195, opened 2023-03-09
Update Rack for Railsbench and Erubi_rails

noahgibbs:
This is meant to obsolete #192 and #193, which have Nokogiri-update bugs common to Dependabot. Smoke-testing now.


noahgibbs:
Smoke test passes, merging.

PR #197, opened 2023-03-12
Tweaks for run_benchmarks.rb

eregon:
Will need to deconflict the `OUT_CSV_PATH` line with https://github.com/Shopify/yjit-bench/pull/194 but should be trivial.
See individual commit messages for the changes.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -354,7 +354,7 @@ OptionParser.new do |opts|
     args.with_pre_init = str
   end
 
-  opts.on("--rss", "measure RSS after benchmark iterations") do
+  opts.on("--rss", "show RSS in the output (measured after benchmark iterations)") do
     args.rss = true
   end
 


k0kubun:
I wonder if we can rename harness directories to something like:

```bash
harness/bips
harness/continuous
harness/perf
harness/stats
harness/warmup
harness/default # renamed from harness
```

so that we can just pass `--harness=warmup` instead of `--harness=harness-warmup`. We're getting more harnesses lately. It would tidy up the repository root too.

eregon:
The downside of that is ruby's `-I` is not enough then to specify the harness.
IMO it's not worth the complexity and I don't see any issue with a few `harness-*` at the root, it's actually to show there is not just one harness.
We could have special handling for the `--harness` option to automatically append `-harness` if not already there though, I can add that.

eregon:
This is now ready for review.
It fixes https://github.com/Shopify/yjit-bench/pull/201/files#r1139215845
And makes this work:
```
$ ruby run_benchmarks.rb --rss --harness warmup -e "ruby::ruby" fib
```
And also the RSS are saved in the JSON file.

benchmarks/30k_methods.rb
```diff
@@ -120009,7 +120009,7 @@ def inc()
 
 require 'harness'
 
-INTERNAL_ITRS = ENV.fetch("INTERNAL_ITRS", 2000).to_i
+INTERNAL_ITRS = Integer ENV.fetch("INTERNAL_ITRS", 2000)
```
k0kubun, line 120012: trivial: I wonder why you sometimes put parenthesis for `Kernel#Integer` and sometimes don't. A method name starting with a capital letter is already confusing, so I personally prefer always put parenthesis. WDYT?
```suggestion
INTERNAL_ITRS = Integer(ENV.fetch("INTERNAL_ITRS", 2000))
```

eregon: I usually add parens for `Integer()`, indeed. I think here I was just trying to minimize the diff and noise, but I'll add parens, it's clearer indeed.

eregon:
As a note, RSS is always measured since #201. So the `--rss` option is just whether to include the RSS measurements in the various outputs. I'll update the option description to clarify that.

PR #199, opened 2023-03-13
Create a microbenchmark for the throw instruction

maximecb:
I thought it would be interesting to have this just so we can compare the performance overhead of the YJIT throw implementation vs the interpreter.

### DIFF
--- a/benchmarks/throw.rb
+++ b/benchmarks/throw.rb
@@ -1,39 +1,30 @@
 require 'harness'
 
 def foo
-  throw 'error'
+  yield
 end
 
-def foo1
-  foo
-end
-
-def foo2
-  foo1
-end
-
-def foo3
-  begin
-    foo2
-  rescue => err
-    err
-  end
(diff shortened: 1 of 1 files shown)


benchmarks/throw.rb
```diff
@@ -0,0 +1,39 @@
+require 'harness'
+
+def foo
+  throw 'error'
```
k0kubun, line 4: This is a different `throw` from what we're talking about. This is an `opt_send_without_block` instruction, not `throw` insn.

If you look at liquid's example code that I shared in https://github.com/Shopify/ruby/issues/504#issuecomment-1462688863 ([example](https://github.com/Shopify/liquid/blob/48cb643c026557f48e524dfd39cc9ff90aa3db95/lib/liquid/context.rb#L247)), it's a `return` inside a block, which cannot be `leave` instruction because it does make a cross-instruction jump from the block to the method.

PR #211, opened 2023-03-24
Remove extensions and plugins from Sequel benchmark

jeremyevans:
These are unrelated to the benchmark itself (other than maybe prepared_statements, which I recommend against using).

Extensions:

* migration: not used (almost never used at runtime in projects using Sequel)
* escaped_like, sql_log_normalizer: not used, not even loaded into the created DB object
* sqlite_json_ops: not used

Plugins:

* association_dependencies, nested_attributes: not used, no associations defined, no code paths exercised as destory is not called
* auto_validations, validation_helpers, require_valid_schema, subclasses: not used inside the benchmark code, no code paths exercised
* prepared_statements: no longer recommended for in Sequel code, only exists for backwards compatibility
* json_serializer: not used, no code paths exercised
* serialization: loaded into superclass, references a column that doesn't even exist in the subclass

The only reason to load these is if you wanted to increase marking time during full garbage collection.

### DIFF
--- a/benchmarks/sequel/benchmark.rb
+++ b/benchmarks/sequel/benchmark.rb
@@ -6,14 +6,6 @@ use_gemfile
 
 require "sequel"
 
-# Sequel with common plugins & extensions
-%i[
-  escaped_like
-  migration
-  sql_log_normalizer 
-  sqlite_json_ops
-].each { |name| Sequel.extension name }
-
 DB = Sequel.sqlite
 
 DB.create_table :posts do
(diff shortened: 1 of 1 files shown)


hmistry:
Jeremy, thank you for the explanation. This makes sense then. Are there other plugins or queries representing real world use cases that can benchmarked to see if and how YJIT optimizes the plugin strategy?

jeremyevans:
Most Sequel extensions and plugins have real world use cases that could be benchmarked.  I was told that the Sequel benchmark should be similar to the ActiveRecord benchmark.  So if additional code is added to the ActiveRecord benchmark, and requires features that Sequel ships in plugins/extensions, then those plugins/extensions should be used in the Sequel benchmark.

Note that I kept the Sequel timestamps plugin for more similarity to ActiveRecord, but it's actually not used in the benchmarked code, and doesn't actually affect behavior during benchmark setup (the Post.create loop), since the timestamp column defaults are both set to CURRENT_TIMESTAMP and the Post instances are never updated after creation.

hmistry:
Jeremy, you make a good point about the code paths that don't get exercised. Benchmarking is tough and unless there isn't a good variety of code flows and data structures exercised, it'll be hard to evaluate JIT optimizations. I find it limiting to have to be same as AR even though I understand it becomes overwhelming to look at.

I'm glad this highlighted a bug or performance optimization in Sequel. Thanks for updating this.

PR #225, opened 2023-04-17
Show interp/yjit at the end

k0kubun:
We are often more interested in `interp/yjit` than `yjit 1st itr` when we look at yjit-bench results. @paracycle said his eyes go to the last column first, it's confusing that it's not the final result, and `interp/yjit` should be in that column. And I felt the same way when I looked at https://bugs.ruby-lang.org/issues/4040#note-21. 

I also think it would be more intuitive if the first result were shown first (on the left) and later results were shown later (on the right).

### Before
```
----------  -----------  ----------  ---------  ----------  -----------  ------------
bench       interp (ms)  stddev (%)  yjit (ms)  stddev (%)  interp/yjit  yjit 1st itr
railsbench  1367.1       0.0         836.7      0.0         1.63         0.36
----------  -----------  ----------  ---------  ----------  -----------  ------------
```

### After
```
----------  -----------  ----------  ---------  ----------  ------------  -----------
bench       interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
railsbench  1367.1       0.0         836.7      0.0         0.36          1.63
----------  -----------  ----------  ---------  ----------  ------------  -----------
```

### DIFF
--- a/misc/graph.rb
+++ b/misc/graph.rb
@@ -37,7 +37,7 @@ def render_graph(csv_path, png_path, title_font_size: 16.0, legend_font_size: 12
   rubies.drop(1).each_with_index do |ruby, index|
     speedup = table.drop(1).map do |row|
       num_rests = rubies.size - 1
-      row.last(num_rests * 2).first(num_rests)[index]
+      row.last(num_rests)[index]
     end
     g.data ruby, speedup.map(&:to_f)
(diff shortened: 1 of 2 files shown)


maximecb:
Good idea

PR #226, opened 2023-04-25
Add `--no-pinning` command-line option to disable CPU core pinning

maximecb:
We already detect TruffleRuby/JRuby and disable CPU core pinning, but for the paper, I feel we should benchmark every Ruby (including YJIT) with no CPU core pinning, just so the configurations are the same for everyone.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -205,7 +205,7 @@ def sort_benchmarks(bench_names)
 end
 
 # Run all the benchmarks and record execution times
-def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_path:, harness:, pre_init:, rss:)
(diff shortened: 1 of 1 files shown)


noahgibbs:
Mostly this looks good. It feels like maybe there should be a pinning option that should either be true or false to override the default? But that's not necessary for our current use-case, so it would be more of a nice-to-have. Other than that, this looks great.

run_benchmarks.rb
```diff
@@ -298,6 +298,7 @@ def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_pat
   name_filters: [],
   rss: false,
   graph: false,
+  no_pinninng: false,
```
eregon, line 301: There is a typo here: no_pinninng -> no_pinning

maximecb: 🤦‍♀️ fixed

Issue #227, opened 2023-04-27
`run_benchmarks.rb` should store all raw data in json output


maximecb:
Currently, we don't store the rss value or the stats reported by `harness-mplr`. Since we already produce a JSON output file, we could store the raw data for all benchmarks in there as well.

@noahgibbs

PR #228, opened 2023-04-27
Store raw data in JSON output file

maximecb:
This will be useful so we can store info such as the rss and stats.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -208,6 +208,7 @@ end
 def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_path:, harness:, pre_init:, rss:, no_pinning:)
   bench_times = {}
   bench_rss = {}
+  bench_data = {}
 
   # Get the list of benchmark files/directories matching name filters
   bench_files = Dir.children('benchmarks').sort.filter do |entry|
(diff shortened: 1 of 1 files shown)


maximecb:
This addresses #227

k0kubun (approved):

eregon:
This means we have the same data twice in the final JSON file (the timings, the largest part):
```
$ ruby run_benchmarks.rb -e "yjit::ruby --yjit" fib
$ jq < ./data/output_038.json
```
```json     
{
  "metadata": {
    "yjit": "ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [x86_64-linux]"
  },
  "raw_data": {
    "yjit": {
      "fib": {
        "RUBY_DESCRIPTION": "ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [x86_64-linux]",
        "values": [
          0.05149682399996891,
          0.050308195999832606,
          0.050372757000104684,
          0.050349385999879814,
          0.050266145000023243,
...
        ],
        "rss": 17461248
      }
    }
  },
  "yjit": {
    "fib": [
      51.49682399996891,
      50.308195999832606,
      50.372757000104684,
      50.349385999879814,
      50.26614500002324,
...
    ]
  }
}
```

I think just keeping the raw_data would be sufficient, and then that JSON by run_benchmarks.rb is just aggregating all individual benchmark runs JSON into one, by Ruby and benchmark.

PR #233, opened 2023-05-16
Upgrade to Nokogiri 1.15.0 for erubi_rails and railsbench

noahgibbs:


casperisfine:
That's only a problem with older rubies. I updated bundler and re-ran bundler update, without any manual edits and it's now green.

casperisfine (approved):

noahgibbs:
Okay. I'll try merging it. If it starts failing benchmark CI again I'll try upgrading stuff there.

PR #240, opened 2023-07-06
Save the size of ruby-lsp responses and verify against expected sizes

### LINKED ISSUE #239: Verify ruby-lsp is working during run
I'm seeing results from @eightbitraptor that look a *lot* like the ruby-lsp benchmark is failing, rapidly, and just not noticing, so it returns really fast, really unstable results.

So I figure I'll add some verification, to ensure that it's generating the right results, for some definition of that phrase. Probably "save the result from the last iteration, do a few basic checks on the value" sort of thing.

noahgibbs:
Presumably the results will be the same on x86 or ARM. If they're not, that's a *really* interesting result...

Fixes #239.

### DIFF
--- a/benchmarks/ruby-lsp/Gemfile
+++ b/benchmarks/ruby-lsp/Gemfile
@@ -3,8 +3,7 @@
 source "https://rubygems.org"
 
 gem "ruby-lsp"
-group :development do
-  gem "rubocop"
-  gem "rubocop-performance"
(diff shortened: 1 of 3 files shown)


noahgibbs:
Interesting. The first call is returning nil on CI. So that's another vote in favour of "sometimes this fails and we don't notice."

noahgibbs:
At this point, I think this is good to go.

PR #243, opened 2023-07-26
Add fluentd benchmark

k0kubun:
This PR adds a new benchmark `fluentd`. Fluentd is a log collector that Shopify currently uses for services like Core and SFR. When Core calls `Rails.logger.info("foo=bar")`, our Fluentd config parses it into `{"foo":"bar"}` and forwards the log to Splunk, for example.

The benchmark code is inspired by the [[JA]How fast really is Ruby 3.x?](https://rubykaigi.org/2022/presentations/fujimotos.html) talk in RubyKaigi 2022. According to him, a Fluentd committer, their parser creates a Ruby object(s) for each log, so the parser is a very runtime-intensive part. Benchmarking the forwarding feature could be challenging in a single-process script, so this benchmarks only the parser. Instead of his synthetic [ltsv.rb](https://github.com/fujimotos/RubyKaigi2022/blob/master/langs/ltsv.rb) that is shown later in his talk, this PR uses the actual [Fluent::Plugin::LabeledTSVParser](https://github.com/fluent/fluentd/blob/v1.16.2/lib/fluent/plugin/parser_ltsv.rb#L38-L49), while parsing the same thing as his [data.ltsv](https://github.com/fujimotos/RubyKaigi2022/blob/472b965c00b3627c4321995793ad235d042f22b9/langs/Makefile#L18).

Since he said "CRuby performance is very important for Fluentd" in the talk, I thought it'd be interesting to include it as a Ruby benchmark.

## Current result

```
ruby: ruby 3.3.0dev (2023-07-26T17:58:15Z master 37160be439) [x86_64-linux]
yjit: ruby 3.3.0dev (2023-07-26T17:58:15Z master 37160be439) +YJIT [x86_64-linux]

-------  ---------  ----------  ---------  ----------  ------------  ---------
bench    ruby (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  ruby/yjit
fluentd  1528.4     0.2         1384.9     0.1         1.10          1.10
-------  ---------  ----------  ---------  ----------  ------------  ---------
```

### DIFF
--- a/benchmarks/fluentd/benchmark.rb
+++ b/benchmarks/fluentd/benchmark.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'harness'
+
+Dir.chdir(__dir__)
+use_gemfile
+
+require 'fluent/engine'
+require 'fluent/parser'
+
+# Prepare a fixture
+ltsv = 1000000.times.map do
+  "time:2022-08-07 07:38:31,842	module:main.py	level:DEBUG	message:No kernel command line url found.\n"
+end.join
+
+# Prepare an LTSV parser
+parser = Fluent::Plugin::LabeledTSVParser.new
+parser.configure(Fluent::Config::Element.new('parse', '', {}, []))
+
+# Benchmark the `<parse>@type ltsv</parse>` use case
+run_benchmark(10) do
+  parser.parse(ltsv) {}
+end
--- a/benchmarks.yml
+++ b/benchmarks.yml
@@ -49,6 +49,8 @@ etanni:
   desc: etanni is an older, extremely simple template-lang format that basically turns your template into an "eval" with a lot of heredocs.
 fannkuchredux:
   desc: fannkuchredux from the Computer Language Benchmarks Game.
+fluentd:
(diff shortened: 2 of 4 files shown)


benchmarks/fluentd/benchmark.rb
```diff
@@ -0,0 +1,23 @@
...
+Dir.chdir(__dir__)
+use_gemfile
+
+require 'fluent/engine'
+require 'fluent/parser'
+
+# Prepare a fixture
+ltsv = 1000000.times.map do
+  "time:2022-08-07 07:38:31,842	module:main.py	level:DEBUG	message:No kernel command line url found.\n"
+end.join
```
maximecb, line 14: Can you add a bit of variety to the messages? Are there different types of log messages?

k0kubun: Looks like this is an error message from `cloud-init`. I used some error messages in its [main.py](https://github.com/number5/cloud-init/blob/a31370afe3887be67bfa2b499e102bcc10814447/cloudinit/cmd/main.py) at https://github.com/Shopify/yjit-bench/pull/243/commits/38d4d0fb49d599d8a8d64d4f9583d48169946138.

PR #246, opened 2023-08-14
Remove assets-related dependencies in Lobsters benchmark

noahgibbs:
I'm not sure why we didn't see this problem earlier, including on literally the same host machine. I'm not seeing it with either yjit-bench or yjit-metrics, even after I remove the relevant gems. So: not sure exactly what's happening here.

I've removed the things that clearly depend on execJS -- I think that should get rid of the error we're currently getting in CI, where it can't find a JS runtime for execJS. We're already not building assets, nor serving them.

### DIFF
--- a/benchmarks/lobsters/Gemfile
+++ b/benchmarks/lobsters/Gemfile
@@ -17,7 +17,7 @@ gem 'sprockets-rails', '2.3.3'
 # js
 gem "jquery-rails", "~> 4.3"
 gem "json"
-gem "uglifier", ">= 1.3.0"
+#gem "uglifier", ">= 1.3.0"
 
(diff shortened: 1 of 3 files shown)


maximecb:
Can I ask why it needs a JS runtime? It's not executing JavaScript code, is it?

noahgibbs:
JS is used during asset builds. It's not executing JS during the benchmark. In fact for this benchmark, we're not even doing an asset build ahead of time. But we *are* running in prod mode, so it tried to make sure that it *could* build assets, even though we aren't using them here.

maximecb:
Thank you for taking the time to do this Noah 🙏

PR #249, opened 2023-09-06
Make -Iharness optional

k0kubun:
You need `-Iharness*` to run a single benchmark file, but in many cases I just want to use the default `-Iharness`. This PR lets benchmark files use `harness/harness.rb` by default and make `-I` optional while still allowing you to change it with `-I`.

This PR adds a script called `harness/setup.rb` (named like `bundler/setup.rb`), which adds `harness/` as a load path with the lowest priority and requires `harness`. `-I` load path takes precedence, so it can overwrite what's loaded there.

### DIFF
--- a/harness/setup.rb
+++ b/harness/setup.rb
@@ -1,0 +1,3 @@
+# Use harness/harness.rb by default. You can change it with -I option.
+$LOAD_PATH << File.expand_path(__dir__)
+require "harness"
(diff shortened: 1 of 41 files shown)


harness/setup.rb
```diff
@@ -0,0 +1,3 @@
+# Use harness/harness.rb by default. You can change it with -I option.
```
maximecb, line 1:
```suggestion
# Use harness/harness.rb by default. You can change it with -I option.
# i.e. ruby -Iharness benchmarks/railsbench/benchmark.rb
```

maximecb (approved):

PR #251, opened 2023-09-18
Add TruffleRuby in CI

eregon:
It would be nice if benchmarks also ran a single iteration on TruffleRuby in CI.
Currently all benchmarks pass after https://github.com/Shopify/yjit-bench/pull/250

If some new benchmark does not work on truffleruby or there is some issue with a benchmark on truffleruby it would be fine to just exclude that benchmark on truffleruby, the goal is not to prevent merging benchmarks which do not work on truffleruby but just raise awareness if a change to yjit-bench does not work on TruffleRuby.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ head, truffleruby-head ]
+        ruby: [ ruby, head, truffleruby-head ]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


eregon:
Actually I think it'd be good to test on the latest CRuby release too, to make sure benchmark run fine there too, I pushed a commit for that.

eregon:
Yes, that's fixed by https://github.com/Shopify/yjit-bench/pull/250, I didn't include these commits again here to make the diff easier to review. Will rebase once that's merged.

PR #253, opened 2023-10-03
Specify a git ref for rails

k0kubun:
I don't know exact steps to reproduce this, but I often get this error when running `lobsters` with Ruby master. 

```
Fetching gem metadata from https://rubygems.org/.........
Your bundle is locked to rails (7.1.0.beta1) from https://github.com/rails/rails.git (at main@d21d811), but that version can no longer be found in that source. That
means the author of rails (7.1.0.beta1) has removed it. You'll need to update your bundle to a version other than rails (7.1.0.beta1) that hasn't been removed in
order to install.
```

This is fixed if you have `ref` in Gemfile. The local git caching doesn't seem to work well unless you specify `ref`. This could be a bug on the rubygems side, but I'd like to run benchmarks whether we fix this or not.

### DIFF
--- a/benchmarks/lobsters/Gemfile
+++ b/benchmarks/lobsters/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "rails", github: "rails/rails"
+gem "rails", github: "rails/rails", ref: "d21d811ffece4d3959bcd37e58fec77590ff6f93"
 
 gem "sqlite3"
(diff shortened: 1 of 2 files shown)


noahgibbs:
Fair. To update Rails without the ref you can do "bundle update rails". But it's also good to do it explicitly like this -- much more trackable.

noahgibbs (approved):

PR #255, opened 2023-10-06
Restructure Ruby LSP benchmarks

vinistock:
As promised in #239, this PR splits the Ruby LSP benchmarks into three, with the most expensive operations we typically perform. I recommend reviewing by commit.

1. Moved the existing benchmark into `ruby-lsp-rubocop`. That benchmark is now only concerned with collecting violation diagnostics using RuboCop
2. Created `ruby-lsp` to check the performance of automatic requests that occur whenever the user types something in the editor
3. Created `ruby-lsp-indexing` to check the performance of our indexing visitor. This is also performance critical for the LSP since we use it on every file in the project and its gem dependencies to collect declarations


benchmarks/ruby-lsp/benchmark.rb
```diff
@@ -5,31 +5,28 @@
...
-  rc = RubyLsp::Requests::Diagnostics.new(file_uri, document).run
-  rc_last = rc.size
-
-  # Running SyntaxTree visitor requests
-  hl = RubyLsp::Requests::SemanticHighlighting.new(
-    document,
-    encoder: RubyLsp::Requests::Support::SemanticTokenEncoder.new,
-  ).run
-  hl_last = hl.data.size
+  executor.execute({
```
noahgibbs, line 26: Is there any way to check what it's returning? We had those problems where it was just not doing what was asked, and it's really good to check at least one iteration's result, if possible.

maximecb:
I'm a bit hesitant because ruby-lsp was our second biggest benchmark, and if we have 3 different ruby-lsp benchmarks, we're unlikely to look at all 3 in detail just because we have 13 headline benchmarks already. If you want/need 3 different benchmarks for your own needs, it can make sense, but maybe we should keep the original ruby-lsp benchmark as is, and then you could add your 3 new benchmarks on top?

Otherwise, if you really do want to just split it into 3, then I would ask, which operation do you most want to us to optimize for? Which one matters most for common use cases that we care about at Shopify? That one should be the headline benchmark.

Finally, there's a question of names. Ideally benchmark names should be short, because they occasionally have to fit on graphs. Could `ruby-lsp-rubocop` simply be `rubocop` ? Or maybe `ruby-lsp`, `lsp-index`, `lsp-cop` ?

eregon:
Maybe rubocop should be its own independent benchmark? Does ruby-lsp use rubocop in a special way or is it in terms of workload similar to running `rubocop` on the command-line?

benchmarks/ruby-lsp/benchmark.rb
vinistock: That's fair, thanks for the context. I'll add a step that asserts the response is what we expect in all benchmarks. Indeed, it's much better to ensure the results are what we expect than getting incorrectly inflated results.

eregon:
@nirvdrum was mentioning ruby-lsp on truffleruby performance issues and this would be a great way to investigate.
@vinistock could you maybe pick this up again when you have time?

vinistock:
Okay, I've changed the previous Ruby LSP benchmark to be RuboCop only. And now the Ruby LSP benchmark verifies our indexing functionality, which is currently the most performance intensive operation we have to perform.

Let me know if you have any more feedback!

maximecb:
Thanks @vinistock 

Just testing the benchmarks locally, I think the only issue is that the `ruby-lsp` benchmark runs too fast (2ms per iteration). It's better if they take more time per iteration to get more accurate timings.

Could you use a bigger source file to test ruby-lsp. Something at least 10x larger. Ideally real-world code and not synthetic. You could potentially use a piece of Ruby code from ruby-lsp's source code, for example? What's the biggest source file you have in that codebase?

Also, rubocop seems to run 5x faster with YJIT? (wat??? how?)

vinistock:
I switched to use the `node.rb` file from Prism, which is bundled in Ruby 3.3.0. That's the slowest file to index considering all files in the Ruby LSP + dependencies, taking something like 50-70ms.

Let me know if it's okay to read the bundled file from Ruby itself.

> Also, rubocop seems to run 5x faster with YJIT? (wat??? how?)

I'm not seeing that. Where/how are you seeing that 5x number? These are the results on my machine.

```
-------  -----------  ----------  ---------  ----------  ------------  -----------
bench    interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
rubocop  104.4        1.4         62.5       4.8         0.46          1.67       
-------  -----------  ----------  ---------  ----------  ------------  -----------
```

vinistock:
> Ruby open source license at the top of the file

Is it necessary given that Prism is licensed under Shopify (same license as yjit-bench)? Happy to put the license there, but I thought it wouldn't be necessary.

maximecb:
May be necessary if it has the Ruby license.

Timings for rubocop match with a release build 👍

maximecb:
Thanks @vinistock !

PR #257, opened 2023-11-21
Burn-in test script for YJIT / Ruby 3.3

maximecb:
Run benchmarks in a loop with many random options across many forked processes to try and detect flaky errors.

### DIFF
--- a/burn_in.rb
+++ b/burn_in.rb
@@ -61,7 +61,7 @@ def run_benchmark(bench_name, logs_path, run_time, ruby_version)
     # Write command executed and output
     out_path = free_file_path(logs_path, "error_#{bench_name}")
     puts "writing output file #{out_path}"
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
I wondered how we can retrospectively discover a core dump path, but maybe we can look at pid (not verified). It should be enough to run this script after `ulimit -c unlimited` and look at `/var/lib/apport/coredump`. At least we'll find _some_ core dumps even if we can't match it.

maximecb:
> I wondered how we can retrospectively discover a core dump path, but maybe we can look at pid (not verified). It should be enough to run this script after `ulimit -c unlimited` and look at `/var/lib/apport/coredump`. At least we'll find _some_ core dumps even if we can't match it.

I'll add code to write the pid in the error logs to make things easier 👍

PR #259, opened 2023-11-28
Bump graphql-c_parser to 1.0.7

peterzhu2118:
The gem has a bug with compaction that was fixed in rmosolgo/graphql-ruby#4641.


maximecb (approved):
Thank you Peter :)

k0kubun (approved):

PR #260, opened 2024-01-05
Add benchmarks from HN blog post

maximecb:
The other day I saw this HN blog post about [benchmarking 20 programming languages](https://news.ycombinator.com/item?id=38850651), in which they had benchmarked Python and Ruby (with YJIT!) among many others. The good news is that Ruby with YJIT greatly outperformed Python. The less good news is that we're still far behind other languages including PHP.

I asked the author for permission to include the Ruby benchmarks that he used. These are synthetic with big loops. I think they could be useful to benchmark changes we make to the register allocator. It would also be interesting to see if we can make performance on the matrix multiplication benchmark go up significantly by implementing floating point addition and multiplication.

```
-------  -----------  ----------  ---------  ----------  ------------  -----------
bench    interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
matmul   3408.0       0.6         1537.8     0.1         2.20          2.22       
nqueens  879.6        0.8         880.1      0.1         1.00          1.00       
sudoku   4054.0       0.1         1574.5     0.1         2.59          2.57       
-------  -----------  ----------  ---------  ----------  ------------  -----------
```

It's also interesting that we see zero speed boost on `nqueens` even though `--yjit-stats` reports zero exits. Suggests there is an exit that we're not catching/counting somewhere @k0kubun @XrXr
```
total_insns_count:       182,080,967
vm_insns_count:          182,071,959
yjit_insns_count:              9,008
ratio_in_yjit:                  0.0%
avg_len_in_yjit:                 9.5
total_exits:                    0
```


maximecb:
@nirvdrum you can probably get a 10-20x speedup on these with TruffleRuby :)

k0kubun (approved):

k0kubun:
@eregon It appears that TruffleRuby no longer works correctly with cool.io https://github.com/Shopify/yjit-bench/actions/runs/7426843741/job/20211367451. It doesn't seem like the problem of yjit-bench but TruffleRuby's. Could you fix this issue?

k0kubun:
> It's also interesting that we see zero speed boost on nqueens even though --yjit-stats reports zero exits. Suggests there is an exit that we're not catching/counting somewhere

`nq_solve` is called only num_itrs times. `run_benchmarks.rb` runs only 25 itrs, so it's called only 25 times. So `nq_solve` is never JIT-ed. I think you want to change how the method is executed in the script.

With `--yjit-call-threshold=1`, `total_exits` is not 0. But the method seems to exit on `opt_ltlt` and that seems to make `ratio_in_yjit` 0.0%, so YJIT doesn't do well on that benchmark even with a low call threshold.

k0kubun:
> exit on opt_ltlt

```
left shift (ltlt) exit reasons:
    amt_changed:          1 (100.0%)
```

Ah, so we expected a left shift to not have a dynamic amount, and this benchmark did. The amount can be as large as 11, so if chain guard more than 11 times and/or fallback to a slower implementation instead of existing there, we wouldn't exit.

noahgibbs (approved):

noahgibbs:
Useful but not mandatory: deterministic solvers like this are *great* for doing a quick check before or after the run_benchmark() loop to make sure we're getting the right result. That way if something starts returning the wrong answer we'll know really quickly.

nixme:
I took a stab at cleaning up them up, which got accepted: https://github.com/attractivechaos/plb2/tree/bbff5fba14b7ad55eac1ec1faa6f3db8798bcc70/src/ruby

[Timings improved for matmul and sudoko](https://github.com/attractivechaos/plb2/blob/bbff5fba14b7ad55eac1ec1faa6f3db8798bcc70/README.md):
- matmul: before: 130.51s, after: 64.95s
- sudoku: before: 52.26s, after: 17.47s

PR #261, opened 2024-01-06
Add --bench and --warmup flags

k0kubun:
I often use `WARMUP_ITRS=2 MIN_BENCH_ITRS=3 MIN_BENCH_TIME=0` for quick speedup check. But it's hard to remember and I always use my shell history to pull it up.

This PR proposes to allow specifying it with `--warmup=2 --bench=3`.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -347,6 +347,16 @@ OptionParser.new do |opts|
     args.harness = v
   end
 
+  opts.on("--warmup=N", "the number of warmup iterations (default: 15)") do |n|
+    ENV["WARMUP_ITRS"] = n
+  end
+
+  opts.on("--bench=N", "the number of benchmark iterations (default: 10)") do |n|
+    ENV["WARMUP_ITRS"] ||= "0"
+    ENV["MIN_BENCH_ITRS"] = n
+    ENV["MIN_BENCH_TIME"] ||= "0"
+  end
+
   opts.on("--yjit_opts=OPT_STRING", "string of command-line options to run YJIT with (ignored if you use -e)") do |str|
     args.yjit_opts=str
   end
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -356,6 +356,12 @@ OptionParser.new do |opts|
     ENV["MIN_BENCH_TIME"] ||= "0"
   end
 
+  opts.on("--once", "benchmarks only 1 iteration with no warmup for the default harness") do
+    ENV["WARMUP_ITRS"] = "0"
+    ENV["MIN_BENCH_ITRS"] = "1"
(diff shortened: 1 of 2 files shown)


run_benchmarks.rb
```diff
@@ -347,6 +347,16 @@ def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_pat
     args.harness = v
   end
 
+  opts.on("--warmup=N", "the number of warmup iterations (default: 15)") do |n|
+    ENV["WARMUP_ITRS"] = n
+  end
+
+  opts.on("--bench=N", "the number of benchmark iterations (default: 10)") do |n|
+    ENV["WARMUP_ITRS"] ||= "0"
```
noahgibbs, line 355: Seems a bit odd to default WARMUP_ITRS to 0, but only if --bench is specified.

noahgibbs: You *do* document it that way in the README, so I assume this is on purpose. And that very often matches my usage pattern. I'm just not sure if this will be expected behaviour.

k0kubun: Okay, I dropped that feature to avoid confusion https://github.com/Shopify/yjit-bench/pull/261/commits/9ddaeae8e0c1a4806dc440eee43c91943d1f73af.

run_benchmarks.rb
```diff
@@ -347,6 +347,15 @@ def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_pat
     args.harness = v
   end
 
+  opts.on("--warmup=N", "the number of warmup iterations (default: 15)") do |n|
+    ENV["WARMUP_ITRS"] = n
+  end
+
+  opts.on("--bench=N", "the number of benchmark iterations (default: 10)") do |n|
```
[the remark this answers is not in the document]
k0kubun:
```suggestion
  opts.on("--bench=N", "the number of benchmark iterations for the default harness (default: 10). Also defaults MIN_BENCH_TIME to 0.") do |n|
```

run_benchmarks.rb
```diff
@@ -347,6 +347,15 @@ def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_pat
...
+    ENV["MIN_BENCH_ITRS"] = n
+    ENV["MIN_BENCH_TIME"] ||= "0"
```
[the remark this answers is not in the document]
k0kubun: Accepted your suggestion at https://github.com/Shopify/yjit-bench/pull/261#discussion_r1445076876.

k0kubun:
> How about a --quick flag as well?

How about calling it `--once` https://github.com/Shopify/yjit-bench/pull/261/commits/eebd59a4b67d56c1b57581f157a5918f37598ac5 (since we use the word "once" in `run_once.sh` as well)?

PR #262, opened 2024-01-06
Avoid failing CI on TruffleRuby

k0kubun:
The current `truffleruby-head` doesn't seem to work well with cool.io. This CI job is supposed to test `./run_benchmarks.rb`, not TruffleRuby. Let's avoid failing the CI until TruffleRuby is fixed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
           WARMUP_ITRS: '1'
           MIN_BENCH_ITRS: '1'
           MIN_BENCH_TIME: '0'
+        continue-on-error: ${{ matrix.ruby == 'truffleruby-head' }}
 
       - name: Test run_once.sh
         run: ./run_once.sh --yjit-stats benchmarks/railsbench/benchmark.rb


noahgibbs (approved):

eregon:
Could we skip the problematic benchmark instead?

I added it in https://github.com/Shopify/yjit-bench/pull/251 so it would at least notice regressions for existing benchmarks.
If we do continue-on-error, nobody will notice any regression.

eregon:
The error seems https://github.com/Shopify/yjit-bench/actions/runs/7064460115/job/19232535306 on `fluentd`.
I will take a look. Weird it regressed.

maximecb (approved):

eregon:
Fix in https://github.com/oracle/truffleruby/pull/3372

k0kubun:
Thanks for fixing it. As of this moment, the `truffleruby-head` on GitHub Actions is still failing, so I added a way to skip only `fluentd` for `truffleruby-head` as you suggested https://github.com/Shopify/yjit-bench/pull/262/commits/23b658e417b2a30d8d854b3792a59da328f4618b.

eregon (approved):
Thanks!

maximecb (approved):

eregon:
Self notes:
* Failure for liquid-c: https://github.com/Shopify/yjit-bench/actions/runs/7453361047/job/20278563105
  * RBASIC->klass in iquid-c
* Failure for lobsters: https://github.com/Shopify/yjit-bench/actions/runs/7453479468/job/20278936097
  * Incorrect header handling causing `struct timespec` to not be defined, probably similar to https://github.com/socketry/io-event/issues/88
  * old stringio using RBASIC->flags

PR #269, opened 2024-01-17
Add blurhash benchmark

tenderlove:
This is a benchmark for a pure Ruby version of the Blurhash gem:

  https://github.com/Gargron/blurhash

I think we're trying to find a benchmark that would measure the performance of `setbyte`, but I'm not 100% convinced this is the right benchmark.  It only calls `setbyte` a handful of times per iteration. I think this benchmark might be more suitable for floating point math benchmarks?

Regardless, I think we should add this benchmark and try to speed it up. The C extension version of this code is 10x faster than when YJIT is enabled:

```
$ ruby -I lib bench.rb
ruby 3.4.0dev (2024-01-12T21:13:42Z master 206388b19e) [arm64-darwin23]
Warming up --------------------------------------
                ruby     1.000 i/100ms
                cext     7.000 i/100ms
Calculating -------------------------------------
                ruby      3.589 (± 0.0%) i/s -     18.000 in   5.015422s
                cext     78.361 (± 0.0%) i/s -    392.000 in   5.002508s
$ ruby --yjit -I lib bench.rb
ruby 3.4.0dev (2024-01-12T21:13:42Z master 206388b19e) +YJIT [arm64-darwin23]
Warming up --------------------------------------
                ruby     1.000 i/100ms
                cext     7.000 i/100ms
Calculating -------------------------------------
                ruby      6.596 (± 0.0%) i/s -     33.000 in   5.002780s
                cext     78.332 (± 0.0%) i/s -    392.000 in   5.004399s
```

### DIFF
--- a/benchmarks.yml
+++ b/benchmarks.yml
@@ -79,7 +79,7 @@ sudoku:
 matmul:
   desc: matrix multiplication benchmark
 blurhash:
-  desc: Blurhash calculations
+  desc: blurhash (blurred preview image) calculation
 
 #
 # MicroBenchmarks


tenderlove:
> The `setbyte` function is only called 28 times. I guess it's only used to encode a short hash at the end, and not actually the whole image?

Ya, as I said in the description I think this is probably not a good benchmark for `setbyte` but rather floating point math. We should probably do some work on floating point math, but I'll look for a better `setbyte` benchmark

maximecb:
No stress, probably best to look for relevant benchmarks, pure-Ruby code rather than benchmarks specifically using setbyte.

PR #272, opened 2024-01-20
Include failed benchmarks in output and continue

rwstauner:
console output:

```
railsbench      1176.6       1.5         636.9      1.8         1.30          1.85
ruby-lsp        0.0          NaN         0.0        NaN         NaN           NaN
sequel          38.0         2.0         30.1       2.3         1.24          1.26
```


maximecb:
I would prefer if it had N/A for the entire row of failed benchmarks.

NaN usually indicates that some numerical error happened.

rwstauner:
Thanks, I appreciate that feedback!

I had originally written up a solution that was more intentionally creating the data in the loop (which I think is what we'll need to do eventually).  I lost my changes and I tried to quickly reimplement and so that I could push something to preserve the WIP.  This version ended up being much simpler to start, as instead of rewriting the data generation it just fakes the data into something that can get past the calculations (which is what produces the NaNs).

I expect there will be more work needed to handle (whatever format is used for) the new empty data (anything that reads these CSVs) so I'll need to do a lot of exploring.

maximecb:
Could maybe do something simple, like not store any data for failing benchmarks, then at the end when reporting the data, if there's a benchmark that was in the list to be benchmarked, but has no associated data, you just report N/A for all the values?

maximecb:
Maybe write the stderr output in a log file?

rwstauner:
Well, yeah if there isn't value in jamming the failure into the data report, we could certainly just add extra output to the console.
Would be a quick way to start, at least!
Ok, I'll pursue that and we can decide if we want to do anything else with it down the road.

rwstauner:
You can see an example of the output in the failed `head` build:
https://github.com/Shopify/yjit-bench/actions/runs/7617311415/job/20746050413?pr=272#step:5:1375
(fixed with #274)

Which raises a question: Would we want these CI tests to fail fast?  Or run as much as they can?

maximecb:
I think it would be a lot more noticeable if the two failed benchmarks appeared in the table, maybe as the first rows, with N/A for all values. Makes it hard not to notice.

maximecb (approved):

run_benchmarks.rb
```diff
@@ -32,21 +32,37 @@ def os
...
 end
 
+# Call system, capture stderr, return that with result.
+def run_benchmark_command(command, env: {})
+  puts(command)
+
+  result = {}
+
+  IO.pipe do |reader, writer|
+    result[:success] = system(env, command, err: writer)
```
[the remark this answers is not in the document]
noahgibbs: Fair enough. It will occasionally be different when we hit unexpected cases -- that is, when we hit different exit codes. But reasonable to not bother with it.

run_benchmarks.rb
```diff
@@ -32,21 +32,37 @@ def os
...
   end
 end
 
+# Call system, capture stderr, return that with result.
+def run_benchmark_command(command, env: {})
+  puts(command)
+
+  result = {}
+
+  IO.pipe do |reader, writer|
```
k0kubun, line 52: So I did https://github.com/Shopify/yjit-bench/pull/280, and it works for `--yjit-stats`. But then I also noticed I sometimes print debugging logs to stderr (not stdout so that it doesn't influence tests, etc.). The capturing seems to make it pretty slow, and you don't get any output since the output is no longer streamed.

Can we do either of the following two options?

1. The stderr output is streamed to the console, and show the captured result at the end again.
2. Just avoid capturing stderr.

To my knowledge, there's no easy way to implement (1). You'd need to spawn a Thread to achieve it like Open3 does. Because of the complexity, I'm personally in favor of (2); simply stop capturing `result[:err]`.

WDYT?

eregon: :+1: to avoid capturing stderr, it's enough to scroll above to find the error(s) and definitely do not want to delay those.

rwstauner: Yeah, with the summary at the end making it obvious that some failed, even if you can't scrollback you can always run individual ones again to see the errors.  Always happy to make it simpler: https://github.com/Shopify/yjit-bench/pull/281

PR #281, opened 2024-01-26
Stream stderr from benchmarks

rwstauner:
Instead of capturing it and reprinting it just let it stream. If you can't see the failure output in your scrollback you can just run the failing benchmarks again to see the errors.


k0kubun (approved):
Thanks!

PR #282, opened 2024-02-05
Remove unused `listen` gem from Rails benchmarks

p8:
The `listen` gem is no longer added to the default Gemfile of new Rails app: https://github.com/rails/rails/pull/42985
It is not used in the benchmarks either, as no `file_watcher` is configured in any app.
This also removes the `ffi` gem which can take some time to build.


p8:
I have signed the CLA!

PR #286, opened 2024-03-14
Add new Protoboeuf decoding benchmark

maximecb:


benchmarks/protoboeuf/benchmark_pb.rb
```diff
@@ -0,0 +1,4888 @@
...
+      @z = 0
+
+            tag = buff.getbyte(index)
+        index += 1
+
+
+    while true
+      if tag == 0x8
+                ## PULL_UINT64
+        @x =       if (byte0 = buff.getbyte(index)) < 0x80
```
eregon, line 36: Is this file generated? The indentation etc looks quite odd.
This looks like manual inlining, is it that beneficial with YJIT to be worth it?

benchmarks/protoboeuf/benchmark_pb.rb
```diff
@@ -0,0 +1,4888 @@
...
+        index += 1
+        byte0
+      else
+        if (byte1 = buff.getbyte(index + 1)) < 0x80
```
eregon, line 40: BTW this logic looks like protobuf varint deserialization, is it faster to have this verbose code compared to this simple loop I wrote in Prism?
https://github.com/ruby/prism/blob/20e768d8d75b78147e9588e0b842dbd847b20cfe/templates/lib/prism/serialize.rb.erb#L185-L199

benchmarks/protoboeuf/benchmark_pb.rb
maximecb: Yes

PR #287, opened 2024-04-01
Centralize Ruby Version to `.ruby-version`

jenshenny:
## What are you trying to accomplish?
The `.ruby-version` file is the ecosystem standard for defining a Ruby version.  This PR adds the `.ruby-version` file, and removes all other references to Ruby in this repository, aligning it with the standard.

Verify that the changes in the PR meets the following requirements or adjust manually to make it compliant:
  - [ ] `.ruby-version` file is present with the correct Ruby version defined
  - [ ] There is no Ruby version/requirement referenced in the `Gemfile` (no lines with `ruby  <some-version>`)
  - [ ] A `Gemfile.lock` is built with the defined Ruby version
  - [ ] There is no `TargetRubyVersion` defined  in `rubocop.yml` 
  - [ ] There is no Ruby argument  present in  `ruby/setup-ruby` Github Actions that do **not**  run on a Ruby matrix (no lines with `ruby-version: “x.x”`)

Please merge this PR if it looks good, this PR will be merged if there isn't any activity after 4 weeks.


k0kubun:
For yjit-bench, we need to constantly use multiple Ruby versions. Can we avoid locking the Ruby version in this repository? I don't want `.ruby-version` to inadvertently switch Ruby from master to 3.3.0 when I enter yjit-bench directory, for example.

PR #298, opened 2024-05-07
Remove Rack::Static from rack benchmark

casperisfine:
It does a call to `Dir.pwd` which end up being 50% of the runtime. I'll see upstream if perhaps this call could be cached, and also perhaps see in RUby why it's this slow (It's a syscall, but even for a syscall it seems very slow).

After:
```

-----  -----------  ----------  ---------  ----------  ------------  -----------
bench  interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
rack   72.9         4.8         52.8       2.4         1.24          1.38       
-----  -----------  ----------  ---------  ----------  ------------  -----------
```

### DIFF
--- a/benchmarks/rack/benchmark.rb
+++ b/benchmarks/rack/benchmark.rb
@@ -15,7 +15,6 @@ stack = Rack::Builder.new do
   use Rack::Deflater
   use Rack::Sendfile
   use Rack::ContentLength
-  use Rack::Static
   map "/ok" do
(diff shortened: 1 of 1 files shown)


casperisfine:
Hum, wait, there's something wrong. That `Dir.pwd` is in the rack initialize. Seems like the benchmark was wrong all along.

PR #299, opened 2024-05-07
rack: Only build the middleware stack once

casperisfine:
There was a bug in the benchmark. The app was a `Rack::Builder` meaning each middleware was re-instantiated on every request which isn't what is worth benchmarking.

We need to call `to_app` to get a final middleware stack that is representative of a production system.

```

-----  -----------  ----------  ---------  ----------  ------------  -----------
bench  interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
rack   22.1         21.7        13.4       26.6        1.40          1.65       
-----  -----------  ----------  ---------  ----------  ------------  -----------
```

cc @maximecb @etiennebarrie

### DIFF
--- a/benchmarks/rack/benchmark.rb
+++ b/benchmarks/rack/benchmark.rb
@@ -29,7 +29,7 @@ stack = Rack::Builder.new do
     end
     run app
   end
-end
+end.to_app
 
 env = Rack::MockRequest::env_for("http://localhost/ok")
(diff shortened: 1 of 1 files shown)


maximecb:
Woopsie! Thanks for digging Jean :)

maximecb (approved):

PR #300, opened 2024-05-09
Active Record: remove the query cache

casperisfine:
Followup: https://github.com/Shopify/yjit-bench/pull/297

The idea was to totally skip the native code that can't be optimized much, but I suppose benchmarking the binding code kinda make sense anyway.

@maximecb @eregon 

```
------------  -----------  ----------  ---------  ----------  ------------  -----------
bench         interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
activerecord  185.3        3.5         85.5       4.8         1.95          2.17       
------------  -----------  ----------  ---------  ----------  ------------  -----------
```

### DIFF
--- a/benchmarks/activerecord/benchmark.rb
+++ b/benchmarks/activerecord/benchmark.rb
@@ -57,8 +57,6 @@ Post.transaction do
   end
 end
 
-Post.connection.enable_query_cache! # Avoid spending time in SQLite.
-
 def run
   posts = Post.includes(:comments).order(id: :asc).limit(100)
   posts.each do |post|


k0kubun:
Don't we use this feature in production? If we do, we might want to make the cache hit ratio closer to an actual production environment, e.g. 1% cache miss and 99% cache hit. If the code to handle the cached case is used for 99% of the time, it's what matters the most.

casperisfine:
> The cache hit rate is likely to be very small for real apps,

That's what I thought a while ago, so I instrumented it in production hoping to make a cache to remove it, but the hit rate was surprisingly high (something like 10%, been a while).

Turns out, when you decouple code it's not uncommon for two pieces of code to need to fetch the same data, so you end up duplicating queries.

However I do agree that query cache hits should ideally not happen at all.

PR #301, opened 2024-05-09
Don't print individual run timings

casperisfine:
It's a lot of output you don't need most of the time but more importantly, when running the script from a TTY it can be relatively slow.


casperisfine:
Hum, what about printing a single `.` for each iteration?

maximecb:
IMO the current approach is pretty good. Hasn't been a problem?

maximecb:
But we could add a `--quieter` option for Jean. Will accept that PR ;)

PR #305, opened 2024-07-03
Use full path for out_path

rwstauner:
Without this using an alternate harness on benchmarks that chdir
will raise because it can't find the dir.


rwstauner:
While this seems sensible, it's only causing issues under an unusual situation, so going to close this for now

PR #308, opened 2024-07-16
Reduce dependency requirements to enable running some benchmarks with miniruby

rwstauner:
- **Embed realtime method to remove dependency on benchmark**
- **Warn and skip max rss if fiddle fails to load**
- **Warn and skip writing file if json fails to load**

Makes the output differ thusly:

```
$ WARMUP_ITRS=1 MIN_BENCH_ITRS=2 MIN_BENCH_TIME=0 ruby ~shopify/yjit-bench/benchmarks/protoboeuf-encode/benchmark.rb
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [arm64-darwin23]
itr #1: 102ms
itr #2: 98ms
itr #3: 98ms
RSS: 51.3MiB
MAXRSS: 52912.0MiB
Writing file /Users/rwstauner/src/github.com/ruby/ruby/data/results-ruby-3.3.0-2024-07-16-132905.json
Average of last 2, non-warmup iters: 98ms

$ WARMUP_ITRS=1 MIN_BENCH_ITRS=2 MIN_BENCH_TIME=0 ./miniruby -r./rbconfig ~shopify/yjit-bench/benchmarks/protoboeuf-encode/benchmark.rb
ruby 3.4.0dev (2024-07-16T19:50:00Z master c083a3ffcd) [arm64-darwin23]
itr #1: 96ms
itr #2: 91ms
itr #3: 91ms
RSS: 54.2MiB
Failed to get max RSS: cannot load such file -- fiddle
Failed to write JSON file: cannot load such file -- json
Average of last 2, non-warmup iters: 91ms
```

### DIFF
--- a/harness/harness-common.rb
+++ b/harness/harness-common.rb
@@ -102,7 +102,12 @@ def return_results(warmup_iterations, bench_iterations)
     yjit_bench_results["yjit_stats"] = RubyVM::YJIT.runtime_stats
   end
 
+  write_json_file(yjit_bench_results)
+end
+
+def write_json_file(yjit_bench_results)
   require "json"
+
   out_path = YB_OUTPUT_FILE
   system('mkdir', '-p', File.dirname(out_path))
 
@@ -110,4 +115,6 @@ def return_results(warmup_iterations, bench_iterations)
   puts "Writing file #{out_path}" unless ENV["RESULT_JSON_PATH"]
 
   File.write(out_path, JSON.pretty_generate(yjit_bench_results))
+rescue LoadError
+  warn "Failed to write JSON file: #{$!.message}"
 end


harness/harness.rb
```diff
@@ -19,14 +18,28 @@
...
+    r0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    yield
+    Process.clock_gettime(Process::CLOCK_MONOTONIC) - r0
+  end
+else
+  require "benchmark"
+
+  def realtime(&block)
+    Benchmark.realtime(&block)
+  end
```
k0kubun, line 32: While it does work with `miniruby -I/path/to/ruby/lib`, I think it's more useful to use `Time.new` to obviate `-I` in this case.

k0kubun: Oh never mind, I was misreading the branches. Since `Process.clock_gettime` does work for miniruby, it seems fine.

k0kubun: `optcarrot/tools/shim.rb` has a fallback using `Time.new`, which I think is for very old Rubies, or `gettimeofday` for mruby. I think we don't care about both of them.

rwstauner: Good to know, thanks!
https://github.com/Shopify/yjit-bench/pull/309

PR #315, opened 2024-08-01
Document how to use the perf(1) harness

XrXr:
The other day we lost ~20 minutes due to this race condition,
so I thought I'd document it and also include a general guide.


maximecb:
> The race is bad, but yes, I recommend running a not-too-small number of iterations to minimize the noise when profiling anyway.

The perf harness could warn or force a higher minimum number of iterations to avoid the race condition?

PR #322, opened 2024-09-06
Fix a leak in Lobsters benchmark

casperisfine:
Profiling the lobsters benchmark I'm seeing a fairly large time spent joining tags in TaggedLogger.

That intruigued me, and by digging I noticed the requests UUIDs kept being appended but never removed.

So for every single request we have one more tag, this is skewing the benchmark a bit, as we spend much more time than we should in logging.

Capture d’écran 2024-09-06 à 17 49 33

### DIFF
--- a/benchmarks/railsbench/benchmark.rb
+++ b/benchmarks/railsbench/benchmark.rb
@@ -29,6 +29,7 @@ run_benchmark(10) do
     unless response_array.first == 200
       raise "HTTP response is #{response_array.first} instead of 200. Is the benchmark app properly set up? See README.md."
     end
+    response_array.last.close # Response might be a Rack::BodyProxy and MUST be closed.
   end
 end
 
--- a/benchmarks/lobsters/benchmark.rb
+++ b/benchmarks/lobsters/benchmark.rb
@@ -47,6 +47,7 @@ run_benchmark(10) do
       puts response_array.inspect
       raise "HTTP status is #{response_array.first} instead of 200 for req #{idx}/#{generator.routes.size}, #{path.inspect}. Is the benchmark app properly set up? See README.md."
     end
+    response_array.last.close # Response might be a Rack::BodyProxy and MUST be closed.
   end
 end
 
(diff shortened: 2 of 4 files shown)


casperisfine:
After: 

Capture d’écran 2024-09-06 à 17 53 08

benchmarks/railsbench/benchmark.rb
```diff
@@ -29,6 +29,7 @@
     unless response_array.first == 200
       raise "HTTP response is #{response_array.first} instead of 200. Is the benchmark app properly set up? See README.md."
     end
+    response_array.last.close # Response might be a Rack::BodyProxy and MUST be closed.
```
k0kubun, line 32: question: Do you mean the rest of the elements in the array may not be `Rack::BodyProxy` and not need to be closed?

casperisfine: The `response_array` is `[status, headers, body]`, only body need to be closed.

k0kubun: Ah, got it. I thought it's an array of responses but it's just a single response. That makes sense.

k0kubun (approved):

benchmarks/railsbench/benchmark.rb
rwstauner: The comment says `might be`.  Could it be anything else?

rwstauner (approved):
there is also `benchmarks/railsbench/slow_methods.rb` that could probably use the same fix

benchmarks/railsbench/benchmark.rb
casperisfine: Yeah, could be an IO, but also something that doesn't respond to `close` at all. But in this case it's always true. If one day it's not, we can do a `respond_to?`.

https://github.com/rack/rack/blob/main/SPEC.rdoc#the-body-

> The Body must either be consumed or returned. The Body is consumed by optionally calling either each or call. Then, if the Body responds to close, it must be called to release any resources associated with the generation of the body. In other words, close must always be called at least once;

PR #325, opened 2024-09-09
Update protoboeuf benchmarks with the latest codegen

rwstauner:
- **Update protoboeuf regenerate scripts**
- **Update protoboeuf generated files**

This does not include `append_as_bytes` so that the benchmarks continue to run on older rubies

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
           - ruby: ruby
           - ruby: head
           - ruby: truffleruby-head
+            skip: protoboeuf-encode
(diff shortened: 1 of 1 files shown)


rwstauner:
Thanks, I was just looking at that.
The error is
```
<internal:core> core/string.rb:73:in `bytesplice': wrong number of arguments (given 5, expected 2..3) (ArgumentError)
```
so apparently protoboeuf stopped working on TruffleRuby back in July: Shopify/protoboeuf#105

k0kubun (approved):

eregon:
I filed an issue to track that: https://github.com/oracle/truffleruby/issues/3656

PR #327, opened 2024-09-11
Re-enable Turbo Boost after running benchmarks

k0kubun:
After yjit-bench disables Turbo Boost, it becomes slower to build Ruby and run tests. To avoid that, I want to re-enable Turbo Boost at exit.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -32,8 +32,8 @@ def os
 end
 
 # Checked system - error or return info if the command fails
-def check_call(command, env: {}, raise_error: true)
-  puts(command)
+def check_call(command, env: {}, raise_error: true, quiet: false)
+  puts("+ #{command}") unless quiet
 
   result = {}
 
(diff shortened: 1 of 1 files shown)


run_benchmarks.rb
```diff
@@ -57,13 +57,21 @@ def have_yjit?(ruby)
...
 
+# Disable Turbo Boost while running benchmarks. Maximize the CPU frequency.
 def set_bench_config(turbo:)
+  # sudo requires the flag '-S' in order to take input from stdin
   if File.exist?('/sys/devices/system/cpu/intel_pstate') # Intel
-    # sudo requires the flag '-S' in order to take input from stdin
-    check_call("sudo -S sh -c 'echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo'") unless turbo || intel_no_turbo?
+    unless intel_no_turbo? || turbo
+      check_call("sudo -S sh -c 'echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo'")
+      at_exit { check_call("sudo -S sh -c 'echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo'", quiet: true) }
```
eregon, line 66: It seems better to not make it quiet here to let the user understand why e.g. the sudo password is asked again

k0kubun: Sounds fair. Feel free to file a PR for it, but so far the 15-minute sudo timeout has been long enough for me to never encounter the situation when running benchmarks locally.

run_benchmarks.rb
```diff
@@ -57,13 +57,21 @@ def have_yjit?(ruby)
...
+    end
+    # Disabling Turbo Boost reduces the CPU frequency, so this should be run after that.
     check_call("sudo -S sh -c 'echo 100 > /sys/devices/system/cpu/intel_pstate/min_perf_pct'") unless intel_perf_100pct?
```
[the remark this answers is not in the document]
k0kubun: I mean you've been fine with it so far, so it should be fine. I just kept the code simple by not introducing what I don't need. I'm personally not interested in lowering power consumption, but if you're, feel free to file a PR to change it.

PR #332, opened 2024-09-26
Enable protoboeuf-encode benchmark on TruffleRuby

andrykonchin:
The benchmark was disabled in #325. Now the issue is fixed (in https://github.com/oracle/truffleruby/pull/3674).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
           - ruby: ruby
           - ruby: head
           - ruby: truffleruby-head
-            skip: protoboeuf-encode sequel
+            skip: sequel
     if: ${{ github.event_name != 'schedule' || github.repository == 'Shopify/yjit-bench' }}
     steps:
       - uses: actions/checkout@v3


k0kubun:
Is `truffleruby-head` not updated yet?

```
Failed benchmarks:
  ruby: protoboeuf-encode
```

k0kubun:
TBH we want to test yjit-bench here instead of the TruffleRuby project. Given how unstable this job has been so far, I think this job should live in `oracle/truffleruby` and not in this repository (or at least not make the CI red when TruffleRuby fails).

andrykonchin:
Yeah, looks like a build for the fix isn't ready yet (the fix was merged today). Will retry jobs tomorrow.

maximecb:
> I think only protoboeuf would fail, which seems OK as it's a very CRuby/YJIT-specific gem, considering the generated code it emits (huge methods).

You should have no problem handling that generated code. It's not CRuby-specific.

eregon:
It executes correctly, it's just that https://github.com/Shopify/yjit-bench/blob/d3ce8dace054c0ac1a3339b69e1d5b686fc5868c/benchmarks/protoboeuf/benchmark_pb.rb#L888 is too big to be compiled by a method JIT (1400 lines long).
Anyway, I shouldn't have mentioned that here, it's pretty much orthogonal.

maximecb:
> too big to be compiled by a method JIT (1400 lines long).

I think you mean your method JIT. Google's V8 handles this sort of thing just fine ;)

eregon:
Actually it compiles fine, my bad, probably because it's a lot of very simple arithmetic:
`[engine] opt done   engine=1  id=2126  ProtoBoeuf::TrunkItem#decode_from                  |Tier 2|Time   236( 110+126 )ms|AST 11185|Inlined  47Y   0N|IR   4401/  7516|CodeSize   29163|Addr 0x7f20e00da000|UTC 2024-09-27T10:10:02.328|Src benchmark_pb.rb:888 0x2acaf309`

In the source code it certainly looks like a lot of duplication and something the JIT should be able to decide whether to inline or not (which it could if it's a separate method).

---

I'll make a PR to switch to a release of TruffleRuby in CI.

PR #333, opened 2024-09-26
Allow TruffleRuby CI to fail without failing the whole job

rwstauner:
refs https://github.com/Shopify/yjit-bench/pull/332#issuecomment-2377505433

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,7 @@ jobs:
           - ruby: head
           - ruby: truffleruby-head
             skip: protoboeuf-encode sequel
+            continue-on-error: true
     if: ${{ github.event_name != 'schedule' || github.repository == 'Shopify/yjit-bench' }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


eregon:
Switching to a TruffleRuby release seems better, because indeed this is completely hiding the fact the job fail and e.g. a breaking change in the harness would be unnoticed.
(https://github.com/eregon/truffleruby-gem-tracker would still notice the failure, but not PRs on this repo)

Alternatively, IMO it's also find to keep things as-is, the job only started to fail when updating protoboeuf in #325.
Skipping the benchmark and pinging me sounds a reasonable workflow to me, that doesn't seem to be much overhead for yjit-bench maintainers, is it?

k0kubun:
> that doesn't seem to be much overhead for yjit-bench maintainers, is it?

To me, it does. Nothing feels like more of an overhead than false positives for the repository.

> Switching to a TruffleRuby release

I'm open to trying this first too. As long as the job doesn't fail unless we bump the TruffleRuby version or we push CRuby-specific code, I'm good.

eregon:
Let's do https://github.com/Shopify/yjit-bench/pull/334 then

PR #337, opened 2024-10-01
lobsters: Avoid continiously instantiating HTMLEntities

k0kubun:
This PR copies the following changes to lobsters in this repository:

* https://github.com/lobsters/lobsters/pull/1320
* https://github.com/lobsters/lobsters/pull/1329 which fixed a bug in the above PR

I expected this to reduce the number of evals coming from this gem, but it didn't. As discussed in the PR, that's just how the gem works. However, this patch still removes unneeded allocations, so I made it a PR anyway.


maximecb (approved):
Thank you Kokubun 🙏  Looks ready to merge?

PR #340, opened 2024-10-06
Store the command line in the JSON file

eregon:
* So the used flags are recorded in the JSON file (some flags do not affect RUBY_DESCRIPTION).
* Print the JSON file path for convenience.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -592,7 +592,6 @@ puts(output_str)
 puts
 puts "Output:"
 puts out_json_path
-puts out_tbl_path
 if args.graph
   require_relative 'misc/graph'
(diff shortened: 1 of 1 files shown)


run_benchmarks.rb
```diff
@@ -590,6 +591,7 @@ def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_pat
 # Print CSV and PNG file names
 puts
 puts "Output:"
+puts out_json_path
 puts out_tbl_path
```
k0kubun, line 595:
```suggestion
```

It seems too verbose to print both. Can we just print the recommended one? Ideally, `out_tbl_path` should be eliminated too.

k0kubun (approved):

PR #342, opened 2024-10-07
Update run_benchmarks.rb to pass the JSON to render_graph

eregon:
Sorry for the many small PRs, just realized this needs to be updated as well.


eregon:
The test works now

PR #344, opened 2024-11-12
Fix value reported by get_maxrss on macos

maximecb:
The value of `ru_maxrss` is in bytes on macos, which contradicts what the manpage says at:
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getrusage.2.html

Apparently, the manpage is wrong, but this information has never been corrected. I can't seem to get the Apple bug reporting software to work and so am unable to report this issue to Apple :(

### DIFF
--- a/harness/harness-common.rb
+++ b/harness/harness-common.rb
@@ -52,7 +52,7 @@ def get_rss
 end
 
 def is_macos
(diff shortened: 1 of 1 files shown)


rwstauner (approved):
Makes the MaxRSS a lot closer to the RSS 👍

PR #349, opened 2024-12-03
Add dumb loop microbenchmark

maximecb:
This is a modified version of a microbenchmark circulated on twitter by "TheDumbTechGuy". It's obviously not representative of most Ruby code, but it does show areas where we could perform much better with better inlining and better quality code generation.

### DIFF
--- a/benchmarks/loops-times.rb
+++ b/benchmarks/loops-times.rb
@@ -1,0 +1,21 @@
+require_relative '../harness/loader'
+
+# Fix these values for determinism
+u = 5
+r = 7
+
+run_benchmark(10) do
+  a = Array.new(10000, 0)
+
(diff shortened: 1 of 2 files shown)


k0kubun (approved):

benchmarks/loops-times.rb
```diff
@@ -0,0 +1,21 @@
...
+# Fix these values for determinism
+u = 5
+r = 7
+
+run_benchmark(10) do
+  a = Array.new(10000, 0)
+
+  4_000.times do |i|
+    4_000.times do |j|
+      a[i] += j % u
```
eregon, line 12: This modulo in the inner loop might likely dominates the benchmark, I'd suggest to use `& 0b111` or so instead.

PR #354, opened 2024-12-20
Improve benchmark stability with cpusets, nice, and disabling of hyper threading

rwstauner:
This adds to `run_benchmarks.rb` several additional linux features which can help to make the benchmark results more reliable:

- set scheduler priority (niceness) high
- setup two cpusets, one for benchmarks and one for everything else to move existing processes off of benchmark cpu
- disable hyper threading for benchmark cpus
- move parent process off of cpu that will be used for benchmarks

You can see this improve the results when running the same ruby version twice:
it increases the number of benchmarks that compare at `1.000` on two different computers:

image

Screenshot_2024-12-17_20-48-08

By putting all this logic into a module we can easily reuse it from `yjit-metrics`.

I think it would make sense to eventually move the existing checks for `turbo` into this module as well.

### DIFF
--- a/misc/benchmark_mode.rb
+++ b/misc/benchmark_mode.rb
@@ -1,0 +1,314 @@
+# frozen_string_literal: true
+
+require "etc"
+require "open3"
+
+module BenchmarkMode
+  BENCH_SET = "bench"
+  OTHER_SET = "other"
+
+  module Helpers
+    extend self
+
+    # Transform a list like "0,2-4,7" into an array of integers.
+    def list_to_ints(list)
+      list
+        &.split(',')
+        &.flat_map { |s| a, b = s.split('-').map(&:to_i); b ? a.upto(b).to_a : [a] }
+    end
+  end
+
+  class << self
+    # Enable all available settings.
+    def engage!(nice: nil)
+      Nice.renice_process_group!(nice || -15, Process.getpgrp)
+
+      return unless bench_cpus = find_bench_cpus
+
+      # Create "bench" cpuset for our task (and forks).
+      bench_tasks = [$$]
+      CpuSet.new(BENCH_SET, cpus: bench_cpus, tasks: bench_tasks, sched_load_balance: '0').create
+
+      # Put all other tasks in "other" cpuset.
+      other_tasks = all_tasks - bench_tasks
+      other_cpus = all_cpus - bench_cpus
+      CpuSet.new(OTHER_SET, cpus: other_cpus, tasks: other_tasks).create
+
+      # Disable the hyper thread siblings of our CPUs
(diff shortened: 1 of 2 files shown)


rwstauner:
For the curious, it looks like it isn't finding any cpuset files in CI, it's only setting the nice value there.

misc/benchmark_mode.rb
```diff
@@ -0,0 +1,367 @@
...
+      list
+        &.split(',')
+        &.flat_map { |s| a, b = s.split('-').map(&:to_i); b ? a.upto(b).to_a : [a] }
+    end
+  end
+
+  class << self
+    # Enable all available settings.
+    def engage!(nice: nil)
+      Nice.renice_process_group!(nice || -15, Process.getpgrp)
```
maximecb, line 24: Should comment on what this does

rwstauner: There are some comments below where the module is defined but yeah, I can add something here, too.

run_benchmarks.rb
```diff
@@ -74,6 +75,13 @@ def set_bench_config(turbo:)
     end
     check_call("sudo -S cpupower frequency-set -g performance") unless performance_governor?
   end
+
+  if os == :linux && ENV["BENCHMARK_MODE"] != '0'
```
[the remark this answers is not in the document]
rwstauner: Sure, we can make this an arg.
It is currently "on" by default, and doesn't do anything if it can't, but I wanted to be able to disable it for comparison.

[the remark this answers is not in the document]
maximecb: Ok makes sense

maximecb:
Looks pretty good to me but I will also tag Alan and Kokubun to review since it's very important that we get benchmarking right.

run_benchmarks.rb
```diff
@@ -11,6 +11,7 @@
 require 'etc'
 require 'yaml'
 require_relative 'misc/stats'
+require_relative 'misc/benchmark_mode'
```
k0kubun, line 14: Unlike `misc/graph.rb` that supports both standalone execution and use as a library or other standalone scripts, these two files (`misc/stats.rb` and `misc/benchmark_mode.rb`) only make sense as a library for `run_benchmarks.rb`. So it feels confusing to me to have them under `misc/`. WDYT about moving `misc/stats.rb` and `misc/benchmark_mode.rb` to `lib/`?

eregon:
Would love if this PR gets merged, I was looking for `--no-sudo` the other day.

rwstauner:
simpler, machine-specific alternative: https://github.com/ruby/ruby-bench/pull/533

PR #357, opened 2025-01-21
Fix stats printing for floats and negative deltas

rwstauner:
This allows tracking ratio_in_yjit which was previously printing as `ratio_in_yjit:      97,01,453,835,082,627`.

```
$ YJIT_BENCH_STATS=ratio_in_yjit ./run_benchmarks.rb --chruby 'ruby-master --yjit-stats=quiet' hexapdf
...
itr:   time ratio_in_yjit
 #1: 2942ms  90.7154624209111
 #2: 2020ms  4.721474016014753
 #3: 1887ms  1.5029409662104598
 #4: 1926ms -0.7787970216063087
 #5: 1936ms  0.7501346345277966
 #6: 1926ms  0.49650225551091864
...
ratio_in_yjit:      97.01453835082627
```

### DIFF
--- a/harness/harness.rb
+++ b/harness/harness.rb
@@ -59,17 +59,8 @@ def run_benchmark(_num_itrs_hint, &block)
 
     yjit_stats&.each do |key, old_value|
       new_value = RubyVM::YJIT.runtime_stats(key)
-
-      # Insert comma separators but only in the whole number portion.
-      diff = (new_value - old_value).to_s.split(".").tap do |a|
-        # Preserve any leading minus sign that may be on the beginning.
-        a[0] = a[0].reverse.scan(/\d{1,3}-?/).join(",").reverse
-        # Add a space when positive so that if there is ever a negative
-        # the first digit will line up.
-        a[0].prepend(" ") unless a[0].start_with?("-")
-      end.join(".")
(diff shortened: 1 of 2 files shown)


harness/harness.rb
```diff
@@ -59,7 +59,16 @@ def run_benchmark(_num_itrs_hint, &block)
...
-      diff = (new_value - old_value).to_s.reverse.scan(/\d{1,3}/).join(",").reverse
+
+      # Insert comma separators but only in the whole number portion.
+      diff = (new_value - old_value).to_s.split(".").tap do |a|
+        # Preserve any leading minus sign that may be on the beginning.
+        a[0] = a[0].reverse.scan(/\d{1,3}-?/).join(",").reverse
+        # Add a space when positive so that if there is ever a negative
+        # the first digit will line up.
+        a[0].prepend(" ") unless a[0].start_with?("-")
+      end.join(".")
```
maximecb, line 70: Might be code to put this in a utility function

rwstauner: Sure.  I hesitated because we've talked before about how additional definitions in the global env can have an affect on the benchmark, but it would allow us to reuse it in the other location

PR #358, opened 2025-01-29
Update lobsters

byroot:
THe initial snapshot was from a few years back, the application changed a lot since then.

I had to run a bunch of migrations and had to fiddle with various sqlite3/msyql incompatibilities. Also some route changes, but with a bit of effort I managed to make it work again.

What is interesting is it seems lobste.rs became almost 3x as slow in these couple years, I haven't yet profiled to know why.

This branch:

```
itr:   time
 #1: 1532ms
 #2: 1013ms
 #3:  910ms
 #4:  852ms
 #5:  851ms
 #6:  901ms
 #7:  857ms
 #8:  886ms
 #9:  875ms
```

main:

```
itr:   time
 #1:  884ms
 #2:  482ms
 #3:  344ms
 #4:  358ms
 #5:  300ms
 #6:  316ms
 #7:  315ms
 #8:  328ms
 #9:  309ms
```

cc @rwstauner 

Also FYI @pushcx, I'll try to let you know if I find the source of this massive slow-down. But perhaps it's just a red herring.


byroot:
Profile after the upgrade: https://share.firefox.dev/4hrxsdY
Profile before: https://share.firefox.dev/3PTZyCG

Observations:

  - `Sqlite3#step` (queries) went from 19% to 25%.
  - There is a brand new `6.7%` spent in `Kernel#caller`, and that's coming from `Prosopite.subscribe` (this is very fishy).

Nothing else very obvious just yet, after that it's all `< 2%`, so I'll have to play seven differences on the flamegraph.

byroot:
> * `Kernel#caller`, and that's coming from `Prosopite.subscribe` (this is very fishy).

Holy cow, it's digesting the entire backtrace to make a unique key for queries: https://github.com/charkost/prosopite/blob/36f7f5a4fa800fcfa1f9bef5cae94c9739b128fc/lib/prosopite.rb#L263-L272

pushcx:
Thanks for pinging me, I'm very curious to hear what you find. Though we don't have APM set up, I think we would've noticed a prod slowdown that big, so Prosopite or another dev/test change seems plausible.

I should mention two in-progress projects, neither of which has merged to `master`:

 * We've started work migrating [away from MariaDB](https://github.com/lobsters/lobsters/issues/539). That issue only talks about postgres, but we've had some sqlite advocacy this week and the existince of this benchmark's partial port is encouraging. So either it's a lot of churn for you (postgres) or a lot less work for you (sqlite).
  * Tomorrow @joeldrapper and I are experimenting [on stream](https://push.cx/stream) with porting some partials to [Phlex](https://phlex.fun). I'm still uncertain about adopting a young gem as a dependency, but it has a lot of things I like. If we entirely migrate from Erb to Phlex that might reduce the codebase's value to you as a Rails benchmark.

maximecb:
Worth profiling and maybe looking at how much code YJIT generates before/after since the change is so massive, also yjit insn executed count.

byroot:
Alright, so a major piece of that was my mistake. I thought the benchmark had caching entirely disabled, but actually no.

If I re-enable the cache like it was before, it's already much closer:

```
itr:   time
 #1: 1101ms
 #2:  655ms
 #3:  483ms
 #4:  481ms
 #5:  485ms
 #6:  484ms
 #7:  460ms
 #8:  531ms
 #9:  477ms
#10:  472ms
#11:  466ms
#12:  458ms
#13:  459ms
#14:  472ms
#15:  465ms
#16:  460ms
#17:  470ms
#18:  473ms
#19:  456ms
#20:  475ms
#21:  486ms
#22:  469ms
#23:  453ms
#24:  471ms
#25:  467ms
```

Now it's only about 50% slower. I'll see if I can figure out where the remaining difference comes from.

byroot:
Found a regression in `psych`: https://github.com/ruby/psych/pull/713

byroot:
Seems like a big regression is the `/rss` route, the new `stories.rss.builder` is much slower than the old `home/rss.erb`, but I guess that's expected, can't beat ERB with Ruby DSLs. https://github.com/lobsters/lobsters/commit/629df028523f888c3f8fd58d1cdb0f9e261068ba

byroot:
Another important change seems to be:

```ruby
get "/threads" => "comments#threads"
```

That became:

```ruby
get "/threads" => "comments#user_threads"
```

And that new action a at least one order of magnitude more costly. But in a way that is legit, because the old action would simply redirect back to `/active`, so I think it was an oversight in the benchmark, as it didn't make much sense to benchmark that.

maximecb:
Thanks for the investigation Jean 🙏 

How does the percentage in YJIT compare before/after?

> Now there's the question of whether we wish to upgrade or not, given that arguably this newer version of lobsters has much less Ruby code in it. On the other hand, the displayed speedup is more in line with what Rails apps seem to experience in production, so perhaps it's actually a good thing? Dunno.

Yeah I'm torn. On the one hand it's nice if the benchmark is bigger. On the other hand we kind of need as much Ruby code as possible to focus our optimization efforts I feel. When YJIT team members try to implement small optimizations, we look at performance differences on lobsters a lot (because it's the biggest benchmark). If the difference is below 1%, it's hard for us to make the call.

byroot:
> * We should be using an in-memory database like sqlite if we aren't already?

We are already.

XrXr (approved):
It looks like it's the mostly the same amount of ruby work, but there is now more non-ruby workload in the mix to make it slower overall. I share the concern about having a weaker signal for small changes, but a 30% speedup stills seems like a good chunk of the workload. Tracking new changes to keep benchmark relevant is probably important (Rails 8 instead of 7, for example).

I vote that we upgrade.

maximecb:
> I vote that we upgrade.

Jean suggested we could just upgrade Rails while keeping the benchmark largely the same and it shouldn't be too much trouble? We're agreeing on going that path?

> I'd like to get rid of nio4r, still, since the bench doesn't use Action Cable. We can do that in a separate PR, though.

👍

PR #365, opened 2025-07-07
Integrate debug counters

luke-gruber:
If we compile ruby with debug counters and use that ruby for running yjit-bench, there's a mess of output due to the ruby subprocesses that get run (ex: `bundle install`). This is because by default, the output is shown at ruby exit. This change adds a new option `--show-debug-counters` that disables all debug counter output for everything except the benchmark scripts. The counters get reset before each benchmark script is run so that the counters during ruby initialization and prelude are not taken into account.

### DIFF
--- a/harness/harness.rb
+++ b/harness/harness.rb
@@ -39,6 +39,7 @@ def run_benchmark(_num_itrs_hint, &block)
   header = "itr:   time"
 
   RubyVM::YJIT.reset_stats! if defined?(RubyVM::YJIT) && RubyVM::YJIT.enabled?
+  RubyVM.reset_debug_counters if defined?(RubyVM.reset_debug_counters)
 
   # If $YJIT_BENCH_STATS is given, print the diff of these stats at each iteration.
   if ENV["YJIT_BENCH_STATS"]
--- a/harness/harness-common.rb
+++ b/harness/harness-common.rb
@@ -153,6 +153,12 @@ def return_results(warmup_iterations, bench_iterations)
   end
 
   write_json_file(yjit_bench_results)
+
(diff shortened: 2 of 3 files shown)


k0kubun:
How about implementing that as a new harness? Do you need to use this on multiple different harnesses? If you create `harness-debug-counters/harness.rb`, you can use it with `--harness debug-counters`.

luke-gruber:
I can make a harness if you prefer, but I was hoping to use this alongside the `vernier` harness.

k0kubun:
Oh ok, it seems fair to implement that on the harness-common then.

harness/harness.rb
```diff
@@ -39,6 +39,7 @@ def run_benchmark(_num_itrs_hint, &block)
   header = "itr:   time"
 
   RubyVM::YJIT.reset_stats! if defined?(RubyVM::YJIT) && RubyVM::YJIT.enabled?
+  RubyVM.reset_debug_counters if defined?(RubyVM.reset_debug_counters)
```
k0kubun, line 42: So are you fine with this not being used by the vernier harness? I think that's what confused me.

k0kubun: Since you edited this file, I assume you tested this feature on this default harness. Given there's some overhead in using debug counters, I wonder if you really want to do both debug counters and vernier at the same time. The vernier profiling result could be off due to the overhead of debug counters, so you might want to do them separately anyway.

luke-gruber: Ah okay I understand your concerns. I think you're right, maybe I'll just create a new harness. I'll update the PR later this week 👍

luke-gruber:
I'm no longer interested in using debug counters for benchmarking, so I'm closing this for now.

PR #370, opened 2025-07-18
Add ZJIT to run_benchmarks.rb

tekknolagi:
This at least will compare against ZJIT if present.

### DIFF
--- a/run_benchmarks.rb
+++ b/run_benchmarks.rb
@@ -57,6 +57,11 @@ def have_yjit?(ruby)
   ruby_version.downcase.include?("yjit")
 end
 
+def have_zjit?(ruby)
+  ruby_version = check_output("#{ruby} -v --zjit", err: File::NULL).strip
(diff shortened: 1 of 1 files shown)


run_benchmarks.rb
```diff
@@ -432,6 +438,9 @@ def run_benchmarks(ruby:, ruby_description:, categories:, name_filters:, out_pat
   if have_yjit?(RbConfig.ruby)
     args.executables["interp"] = [RbConfig.ruby]
     args.executables["yjit"] = [RbConfig.ruby, "--yjit", *args.yjit_opts.shellsplit]
+  elsif have_zjit?(RbConfig.ruby)
```
rwstauner, line 441: Do we want zjit checked first, or yjit first?
Are we generally building rubies with both, or just one of them?

k0kubun (approved):

rwstauner (approved):

PR #371, opened 2025-07-31
Update Bundler to 2.8.0.dev

k0kubun:
This PR updates `BUNDLED WITH` in Gemfile.lock to 2.8.0.dev as a workaround for https://github.com/Shopify/ruby/issues/622. It also bumps the gem versions not locked in Gemfile to ease this work. I pinned hexapdf and ruby-lsp versions since upgrading them break the benchmark though.

When you run the current yjit-bench with Ruby master, Bundler currently re-execs itself and drops JIT options. I don't know how yjit-metrics works around it, but I have no way to locally run YJIT or ZJIT with benchmarks that have Gemfile.lock. Because the repro condition is limited to Gemfile.lock having non-`2.8.0.dev` version in BUNDLED WITH for Ruby master, this PR uses `2.8.0.dev` as BUNDLED WITH to fix the problem. I confirmed that Ruby 3.4.5 works with this Gemfile.lock.


XrXr (approved):
Makes sense. This is how I worked around the re-exec issues I saw. Thanks for investigating the filing a rubygems issue 🙏🏼

tekknolagi:
Small nit: is there a way to upgrade bundler without upgrading packages? That might change benchmark results in weird and subtle ways

k0kubun:
Filed https://github.com/Shopify/yjit-metrics/pull/417 to explain the potential impact on benchmarks.

PR #372, opened 2025-08-04
Delay loading securerandom until after bundler setup

rwstauner:
On ruby 3.x you can get an error like this if you run this without
securerandom:0.4.1 (required by the lockfile) installed:

> in `check_for_activated_spec!': You have already activated securerandom 0.3.1, but your Gemfile requires securerandom 0.4.1. Since securerandom is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports securerandom as a default gem. (Gem::LoadError)

### DIFF
--- a/benchmarks/shipit/benchmark.rb
+++ b/benchmarks/shipit/benchmark.rb
@@ -1,14 +1,14 @@
-require 'securerandom'
+require_relative '../../harness/loader'
 
 ENV['RAILS_ENV'] ||= 'production'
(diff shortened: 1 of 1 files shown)


tekknolagi (approved):
does the relative positioning of the env var section and the harness loader require matter?

rwstauner:
No we just need any gems to be loaded after the use_gemfile call.
>
> Message ID: ***@***.***>
>
I didn't need to move the loader, I just did it because the env vars
weren't influencing that and it seemed more consistent to have it first.

PR #374, opened 2025-08-08
WIP: Ractor harness

casperisfine:
Some of the simpler benchmarks should be relatively easy to run in ractors, and should be mostly perfectly parallelisable, so it could be interesting to try to run them concurrently in a handful of ractors to see if they perform about as well in these conditions.

So far I only tried to run the `binarytree` benchmark, and it seems it's contending on something (GC?)

```
$ ruby --yjit benchmark.rb 
ruby 3.5.0dev (2025-07-23T11:23:38Z master 85221800ca) +YJIT +PRISM [arm64-darwin24]
itr:   time
 #1:  115ms
 #2:  110ms
 #3:  102ms
 #4:  105ms
...
$ RACTOR_PARALLEL=4 ruby --yjit benchmark.rb 
ruby 3.5.0dev (2025-07-23T11:23:38Z master 85221800ca) +YJIT +PRISM [arm64-darwin24]
itr:   time
 #1:  335ms
 #1:  340ms
 #1:  346ms
 #1:  362ms
 #2:  292ms
 #2:  314ms
```

image

### DIFF
--- a/benchmarks/binarytrees/benchmark.rb
+++ b/benchmarks/binarytrees/benchmark.rb
@@ -46,6 +46,4 @@ end
 
 require_relative '../../harness/loader'
 
-run_benchmark(60) do
-  Btree.bench
-end
+run_benchmark(60, Btree)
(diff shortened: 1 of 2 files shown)


luke-gruber:
I'm going to close PR as the other ractor harness PR is the one we are going with.

PR #375, opened 2025-08-08
[WIP] Ractor harness

luke-gruber:
I was also working on a ractor harness at the same time as Jean 😄 

We can incorporate ideas from both imo.

### DIFF
--- a/harness-ractor/harness.rb
+++ b/harness-ractor/harness.rb
@@ -1,0 +1,71 @@
+require_relative '../harness/harness-common'
+
+Warning[:experimental] = false
+ENV["YJIT_BENCH_RACTOR_HARNESS"] = "1"
+
+RACTORS = [
+  0, # without ractor
+  1, 2, 4, 6, 8, 12, 16, 32
+].freeze
+
+unless Ractor.method_defined?(:join)
+  class Ractor
+    def join
+      take
+      self
+    end
+    alias value take
+  end
+end
+
+def run_benchmark(num_itrs_hint, ractor_args: [], &block)
+  warmup_itrs = Integer(ENV.fetch('WARMUP_ITRS', 10))
+  bench_itrs = Integer(ENV.fetch('MIN_BENCH_ITRS', num_itrs_hint))
+  # { num_ractors => [itr_in_ms, ...] }
+  stats = Hash.new { |h,k| h[k] = [] }
+
+  header = "r:   itr:   time"
+  puts header
+
+  i = 0
+  while i < warmup_itrs
+    args = if ractor_args.empty?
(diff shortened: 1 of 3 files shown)


eregon:
Regarding OptCarrot, have you seen https://github.com/mame/optcarrot/blob/master/bin/optcarrot-bench-parallel-on-ractor ?

That seems like a nice way to make OptCarrot Ractor-compatible without copying many files.
That specific file could be modified to not run the benchmark (e.g. if `$0 != __FILE__`) and required by some other file which could use the new Ractor harness here.

I think it's best to avoid large copies in the repo, and `benchmarks-ractor/optcarrot` seems the only offender in this PR.

k0kubun:
Looks good to me.

> Make 30k_methods benchmark Ractor compatible

What was the change? The diff is so large that GitHub doesn't show the diff.

> Regarding OptCarrot, have you seen https://github.com/mame/optcarrot/blob/master/bin/optcarrot-bench-parallel-on-ractor ?

If this is possible, I guess we could let `benchmarks-ractor/optcarrot/benchmark.rb` load optcarrot from `../../benchmarks/optcarrot`, use shims like `optcarrot-bench-parallel-on-ractor`, and run the benchmark.

eightbitraptor:
Ok. Branch has been updated with revised Optcarrot Ractor benchmark. Thanks all.

harness-ractor/harness.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
+require_relative '../harness/harness-common'
+
+Warning[:experimental] = false
+ENV["YJIT_BENCH_RACTOR_HARNESS"] = "1"
```
eregon: Since this is a new env variable, how about following the new name of this repo? (it will be harder to change it later, but it's easy now)
```suggestion
ENV["RUBY_BENCH_RACTOR_HARNESS"] = "1"
```
And the same for `YJIT_BENCH_RACTORS` below.

eregon:
I gave it a try and made `benchmarks/optcarrot/benchmark.rb` Ractor-compatible in #422, which means no copy at all of OptCarrot files. I checked locally and it runs fine on both ruby-master and 3.4.7.
IOW, optcarrot is treated like other benchmarks (e.g. rubykon) which simply use `if ENV["YJIT_BENCH_RACTOR_HARNESS"]` to make them Ractor-compatible.

The only "copied" benchmark is then knucleotide, and that seems fine as-is because it's pretty small and has significant differences with the fork-based version in `benchmarks/knucleotide/benchmark.rb`.

eregon (approved):

PR #377, opened 2025-08-18
Upgrade remaining benchmarks to Rails 8 

byroot:
Reopening because stupid CLA bot...

Fix: https://github.com/Shopify/yjit-bench/pull/373

Some benchmarks already have been upgrade, this bring the remaining ones to Rails 8.0.2 with load_defaults 8.0.

### DIFF
--- a/benchmarks/lobsters/config/application.rb
+++ b/benchmarks/lobsters/config/application.rb
@@ -21,7 +21,7 @@ Bundler.require(*Rails.groups)
 module Lobsters
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 8.0
+    config.load_defaults "#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}"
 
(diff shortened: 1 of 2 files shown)


benchmarks/activerecord/Gemfile.lock
```diff
@@ -32,23 +33,25 @@ GEM
...
+  activerecord (~> 8.0)
   activerecord-jdbcsqlite3-adapter (~> 70)
   base64
   bigdecimal
   mutex_m
-  sqlite3 (~> 1.4)
+  sqlite3
 
 BUNDLED WITH
-   2.8.0.dev
```
k0kubun, line 54: Please avoid downgrading the Bundler version https://github.com/Shopify/yjit-bench/pull/371. The Bundler bug has not been fixed yet.

k0kubun: Fixed in https://github.com/Shopify/yjit-bench/pull/378

PR #407, opened 2025-09-23
lobsters: Downgrade net-imap to v0.5.6

k0kubun:
https://github.com/ruby/ruby/pull/13926 broke `lobsters` benchmark.

Until https://github.com/ruby/net-imap/pull/525 is released, we should use a net-imap version that doesn't use `Ractor.make_sharable`. We could use GitHub as a source, but I didn't want to slow down `bundle install` for a gem that's probably not used by the benchmark code.


nevans:
FYI: ruby/net-imap#525 was released in v0.5.11.

PR #421, opened 2025-10-17
Add microbenchmarks for more kinds of sends

tekknolagi:


benchmarks/send_bmethod.rb
```diff
@@ -0,0 +1,12 @@
+require_relative '../harness/loader'
+
+# Call with and without args
+define_method(:zero) { :b }
+define_method(:one) { |arg| arg }
+
+run_benchmark(500) do
+  2_000_000.times do |i|
+    zero
+    one 123
```
tenderworks: Should we manually unroll this loop somewhat?  I'm kind of concerned this benchmark will spend a significant amount of time dealing with the block entry (for the `times` method) rather than exercising the `define_method` method calls.

tekknolagi: 6254610942b94f4c0a85e3c7022256139b731bc6

PR #427, opened 2025-11-04
Use the released version of strscan on lobsters and shipit

k0kubun:
https://github.com/ruby/strscan/pull/168 broke lobsters and shipit https://github.com/ruby/ruby/actions/runs/19080798257/job/54508537191. The latest tzinfo.gem still uses the method removed in the PR.

Until we fix it on tzinfo or revert the strscan change, I want to switch to a released strscan to avoid failing the ZJIT ruby-bench CI in the meantime.


tekknolagi (approved):
thank you

k0kubun:
I realized it wasn't an obsolete method, so I fixed the upstream instead https://github.com/ruby/strscan/pull/169. Closing.

PR #431, opened 2025-11-14
Add benchmarks for various Addressable URI operations

rafaelfranca:
We want to track the performance of common operations in the Addressable gem, such as parsing, normalization, component access, and modification.

We heavily use the Addressable gem in our projects, so having benchmarks will help us monitor performance changes over time and ensure that updates to the gem or our codebase do not introduce regressions.

### DIFF
--- a/benchmarks/addressable-getters/benchmark.rb
+++ b/benchmarks/addressable-getters/benchmark.rb
@@ -8,7 +8,7 @@ require "addressable/uri"
 COMPLEX_URI = "https://user:pass@example.com:8080/path/to/resource?query=value&foo=bar#fragment"
 
 run_benchmark(100) do
-  100.times do
+  10000.times do
     # Component access
     uri = Addressable::URI.parse(COMPLEX_URI)
     uri.scheme
--- a/benchmarks/addressable-equality/benchmark.rb
+++ b/benchmarks/addressable-equality/benchmark.rb
@@ -6,7 +6,7 @@ use_gemfile
 require "addressable/uri"
 
 run_benchmark(100) do
-  100.times do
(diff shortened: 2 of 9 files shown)


benchmarks/addressable-getters/benchmark.rb
```diff
@@ -0,0 +1,21 @@
...
+
+Dir.chdir __dir__
+use_gemfile
+
+require "addressable/uri"
+
+COMPLEX_URI = "https://user:pass@example.com:8080/path/to/resource?query=value&foo=bar#fragment"
+
+run_benchmark(100) do
+  100.times do
```
tekknolagi: I think we should a) bump the iteration count to ~10k per loop and b) use while loops because that has less per-loop overhead per than invokeblock and can more directly measure this otherwise quite small thing

tekknolagi (approved):

benchmarks/addressable-equality/Gemfile.lock
```diff
@@ -0,0 +1,16 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    addressable (2.8.7)
+      public_suffix (>= 2.0.2, < 7.0)
+    public_suffix (6.0.2)
+
+PLATFORMS
+  aarch64-linux
```
eregon: I guess that's more of a Bundler thing/issue, but I wonder why the architecture is even included here given those are pure-Ruby gems.

rafaelfranca: AFAIK, that is the default behavior of bundler. It always include the architecture where the gemfile was created

PR #433, opened 2025-11-17
Add benchmark for setting Struct fields

eregon:
* Based on benchmarks/structaref.rb

Ref: https://github.com/ruby/ruby/pull/15055

### DIFF
--- a/benchmarks/structaset.rb
+++ b/benchmarks/structaset.rb
@@ -1,0 +1,28 @@
+require_relative '../harness/loader'
+
+TheClass = Struct.new(:v0, :v1, :v2, :levar)
+
+def set_value_loop obj
+  # 1M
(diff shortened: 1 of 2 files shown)


benchmarks/structaset.rb
```diff
@@ -0,0 +1,31 @@
+require_relative '../harness/loader'
+
+TheClass = Struct.new(:v0, :v1, :v2, :levar)
+
+def set_value_loop obj
+  sum = 0
```
tekknolagi: looks like you are not using sum

k0kubun (approved):

benchmarks/structaset.rb
eregon: Indeed, removed

PR #434, opened 2025-11-18
Extract BenchmarkFilter class to its own file

rafaelfranca:
This make easier to change the filter logic.

I'm working on reorganizing this project to allow multiple benchmarks per libraries in the same folder, but this code isn't easy to change, do I'm working my way down the code to make my change and future ones easier.


eregon:
> allow multiple benchmarks per libraries in the same folder

Some thoughts on this: maybe we could allow multiple `run_benchmark` calls per file and so have those benchmarks in the same file, but that has two problems: one benchmark might have effect on the next (and potentially make it less stable due to having e.g. a bunch of extra objects in the heap, polluted inline caches, etc), and it becomes hard to run a single benchmark.
So yeah, I think that feature would be nice.
One alternative I can think of would be to keep the file structure mostly as-is, but change `category` (per benchmark) to `tags` and then e.g. have an `addressable` tag, so we can have multiple overlapping "groups of benchmarks".

PR #436, opened 2025-11-20
Extract parameters parser to its own object

rafaelfranca:
Extracting the parameter parsing to another object to make easier to test and make the code in `run_benchmark.rb` easier to understand.


tekknolagi (approved):
seems reasonable

PR #437, opened 2025-11-20
Silence warnings created by tests

rwstauner:


test/benchmark_runner_test.rb
```diff
@@ -91,12 +91,18 @@
     end
 
     it 'exits when file does not exist' do
-      assert_raises(SystemExit) { BenchmarkRunner.expand_pre_init('/nonexistent/file.rb') }
+      out = capture_io do
+        assert_raises(SystemExit) { BenchmarkRunner.expand_pre_init('/nonexistent/file.rb') }
+      end
+      assert_includes out, "--with-pre-init called with non-existent file!\n"
```
rwstauner: I don't really care if we assert these or just ignore them but it's a pet peeve of mine to have the test output broken up by uncaught warnings (especially when they are properly induced by the test code).

rafaelfranca (approved):
Those tests are later deleted in my branch, that is why I didn't fix them, but I'm ok with this change.

PR #444, opened 2025-11-21
Move all code in run_benchmarks.rb to CLI.run method

rafaelfranca:
Last refactor, organizing the benchmark run code in a CLI.

Unit tests for this new class are added, and the coverage of this code is now close to 100%.


lib/benchmark_runner.rb
```diff
@@ -80,15 +80,20 @@ def render_graph(json_path)
 
     # Checked system - error or return info if the command fails
     def check_call(command, env: {}, raise_error: true, quiet: false)
+      quiet ||= ENV['BENCHMARK_QUIET'] == '1'
```
k0kubun:
```suggestion
    def check_call(command, env: {}, raise_error: true, quiet: ENV['BENCHMARK_QUIET'] == '1')
```

was this not enough?

rafaelfranca: Yes!

PR #445, opened 2025-11-21
Allow multiple benchmarks per folder in ruby-bench

rafaelfranca:
Today, benchmarks are organized, one per folder. For libraries like addressable, we have 9 folders with repeated Gemfile and Gemfile.lock.

This PR, adds the possibility of adding benchmarks for the same library into the same folder, each with a different name.


README.md
```diff
@@ -101,6 +101,33 @@ It's also convenient for profiling, debugging, etc, especially since all benchma
...
+   benchmarks/erubi/
+     benchmark.rb            # Benchmark name: "erubi"
+     Gemfile
+   ```
+
+3. **Multiple benchmarks per directory** - For related benchmarks sharing dependencies:
+   ```
+   benchmarks/addressable/
+     benchmark-equality.rb   # Benchmark name: "addressable-equality"
+     benchmark-join.rb       # Benchmark name: "addressable-join"
```
k0kubun: This seems like a fair naming convention, but it's a bit harder than now to come up with the benchmark name "addressable-join" (which I need to use when I want to run it with `run_benchmarks.rb`) when I look at the file name `benchmarks/addressable/benchmark-join.rb` because "benchmark-" is in the middle of them.

Do you think it'd be problematic to use `benchmarks/addressable/join.rb` for this case? Like the `benchmarks/xxx.rb` files, I don't think we necessarily need to keep `benchmark` in the file names, and we usually don't have non-benchmark `.rb` files in a benchmark subdirectory anyway.

k0kubun:
I wonder if we need any change for https://github.com/rubybench/rubybench and/or https://github.com/rubybench/rubybench.github.io to support this. cc: @eightbitraptor

README.md
rafaelfranca: > Do you think it'd be problematic to use `benchmarks/addressable/join.rb` for this case?

Not really. I will change to that. It is better.

eightbitraptor (approved):
This looks good to me. Thanks 👍

PR #448, opened 2025-11-22
Add --excludes to run_benchmarks.rb to easily exclude benchmarks

eregon:
`--category` and `--name_filters` don't give a way to exclude some benchmarks.
`--name_filters` seems already quite complicated so adding some negated filter there seems undesirable.
Therefore I think it's best to add an explicit `--excludes=x,y,z` option.

Thanks to the recent refactor, this is pretty easy :)

There are many possible reasons to exclude a given benchmark notably too unstable performance, known to fail on some Ruby version and no point to even `bundle install`, etc.
This is vastly more convenient than having to manually exclude by including every other benchmark by name, especially as more benchmarks are added.

### DIFF
--- a/lib/argument_parser.rb
+++ b/lib/argument_parser.rb
@@ -11,6 +11,7 @@ class ArgumentParser
     :yjit_opts,
     :categories,
     :name_filters,
+    :excludes,
     :rss,
     :graph,
     :no_pinning,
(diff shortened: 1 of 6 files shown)


eightbitraptor (approved):
I think this is a great idea.

PR #456, opened 2025-12-10
Upgrade benchmarks to Rails 8.1

byroot:
For Shipit we'll have to wait on https://github.com/Shopify/shipit-engine/pull/1443


k0kubun (approved):
:+1:

PR #507, opened 2026-04-06
Lobsters: Switch from commonmarker to markly

jhawthorn:
Commonmarker switched to Rust via Magnus which is perpetually incompatible with Ruby HEAD due to baking in the ABI and hardcoding other Ruby internals.

markly is a maintained fork of commonmarker 0.x by @ioquatix (thank you!), which is not using deprecated APIs and we can expect to remain compatible with Ruby HEAD. It's very nearly a drop in replacement.

cc @nobu


XrXr (approved):
Thank you!

PR #510, opened 2026-05-22
Bump bundler to 4.0.12

peterzhu2118:
Bundler 4.0.12 has https://github.com/ruby/rubygems/pull/9529 which fixes the following error on Ruby head when running `bundle install`:

    NameError: uninitialized constant Pathname::SEPARATOR_PAT


k0kubun (approved):
:+1:

PR #514, opened 2026-05-27
Use `sudo --non-interactive` in at_exit to not hang

alanwu-shopify-inc:
I noticed that ruby-bench hangs after a long enough benchmarking session. Turns out it was sudo trying to ask for with `quiet: true`. Since it may hang trying to ask for input during exit, let's use `--non-interactive` to opportunistically re-enable turbo boost.

### DIFF
--- a/lib/cpu_config.rb
+++ b/lib/cpu_config.rb
@@ -52,6 +52,17 @@ class CPUConfig
   def check_pstate(turbo:)
     # Override in subclasses
   end
+
+  # A quiet sudo --non-interactive call that on failure, retries asking for input
+  def sudo_prefer_quiet(command)
+    # For compatibility, we use short flags older sudo versions understand.
+    # -n is --non-interactive, -S is --stdin
+    result = BenchmarkRunner.check_call("sudo -n #{command}", quiet: true, raise_error: false)
+    unless result[:success]
+      result = BenchmarkRunner.check_call("sudo -S #{command}")
+    end
+    result
+  end
 end
 
 # Intel CPU configuration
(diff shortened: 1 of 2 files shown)


k0kubun:
It would leave turbo boost in a random state, depending on how long it took to run benchmarks, and run_benchmarks.rb would exit abnormally.

At this point, can we just drop the `at_exit` hook? I prefer a deterministic behavior and not failing the command for re-enabling turbo boost.

alanwu-shopify-inc:
> It would leave turbo boost in a random state, depending on how long it took to run benchmarks, and run_benchmarks.rb would exit abnormally.

All that is not new with this change, just that before this change it would also hang before exiting abnormally

> At this point, can we just drop the at_exit hook? I prefer a deterministic behavior and not failing the command for re-enabling turbo boost.

It seems still useful for short runs, though. It's also nice to not have to figure out the command to restore turbo boost manually, or have a script ready that does it.

alanwu-shopify-inc:
Related to this, maybe we want to flip it and have `--no-sudo` by default. I think the at_exit hook is important if it turns off turbo boost etc by default.

eregon:
What's the reason to use `quiet: true` BTW? That seems the source of "it hangs and not clear why".

Though I agree hanging/waiting for input after a benchmark run is done is not great.

Maybe a solution would be to spawn a subprocess that runs under sudo, so it keeps root permissions during the whole run and restores the state without needing to ask for the sudo password again.

alanwu-shopify-inc:
I'm not sure why it uses `quiet` either. Dropping `quiet` seems another viable path. It's ugly, but probably not good to hide a sudo call either.

k0kubun:
> > It would leave turbo boost in a random state, depending on how long it took to run benchmarks, and run_benchmarks.rb would exit abnormally.
>
> All that is not new with this change, just that before this change it would also hang before exiting abnormally

I'm not sure if I follow. It would not randomly give up on restoring the state or exit abnormally after the password input. 

> What's the reason to use quiet: true BTW? 

`quiet: true` suppresses two things:

1. `out: File::NULL, err: File::NULL`
2. `puts("+ #{command}")`

I think we want to drop (1) from the sudo command, whether we use `--non-interactive` or not. If we decide to use `--non-interactive`, exiting abnormally without printing `sudo: a password is required` seems pretty bad.

On the other hand, I don't want to see (2) for this sudo command in most cases. However, if we can print it only when (we don't use `--non-interactive` and) it asks for password, it might be useful.

alanwu-shopify-inc:
How's [38dab1e](https://github.com/ruby/ruby-bench/pull/514/commits/38dab1ebcc4056b8fe1e12934f12a06c389c8132)? Basically, try `--non-interactive` first, and if it fails, retry with `quiet: false` and `--stdin`. This way there is no prompt in the happy path.

k0kubun (approved):
sounds good!

PR #515, opened 2026-06-10
Improve RSS measurement

eightbitraptor:
Previously we would just capture the RSS (using either `smaps_rollup` if available, `ps -o` if not), but we would only do this once at the end of each benchmark run.

This change samples the RSS during each benchmark iteration (outside of the timed window), so that we can measure and report on the RSS variance during benchmark runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,11 +284,21 @@ after each iteration with the default harness.
 
 ## Measuring memory usage
 
-`--rss` option of `run_benchmarks.rb` allows you to measure RSS after benchmark iterations.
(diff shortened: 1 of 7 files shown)


k0kubun:
> Fix silent hang restoring CPU turbo at exit

Can we skip the diff in this PR? There's already a PR https://github.com/ruby/ruby-bench/pull/514 for that problem.

k0kubun:
The other commit looks good to me :+1:

eregon (approved):
Nice!

PR #525, opened 2026-07-14
Add Liquid IL benchmark

tekknolagi:
https://github.com/tobi/liquid-il compiles Liquid code to Ruby code. Let's see how well YJIT and ZJIT can optimize that Ruby code.


benchmarks/liquid-il/benchmark.rb
```diff
@@ -0,0 +1,31 @@
...
+  [mod, assigns]
+end
+
+# Sanity: every generated module must produce output, or the timing is empty.
+CASES.each { |mod, assigns| raise "empty render for #{mod}" if mod.render(assigns).to_s.empty? }
+
+run_benchmark(150) do
+  # Each render is quick; render the whole template set several times per
+  # iteration to reduce time-measurement noise (mirrors liquid-render).
+  100.times do
```
k0kubun: Can you consider iterating more per iteration? Like 1000?

This seems to take the shortest time per iteration (27ms) out of all headline benchmarks, so unless we change it, I would expect it to suffer from time-measurement noise more than other benchmarks.

```
interp: ruby 4.1.0dev (2026-07-13T13:26:16Z :detached: cb2ea293c6) +PRISM [x86_64-linux]
yjit: ruby 4.1.0dev (2026-07-13T13:26:16Z :detached: cb2ea293c6) +YJIT +PRISM [x86_64-linux]

----------------------  -------------  -------------  ------------  -----------
bench                     interp (ms)      yjit (ms)  yjit 1st itr  interp/yjit
activerecord             339.0 ± 0.0%   157.9 ± 0.0%         2.091        2.146
chunky-png               948.6 ± 0.0%   404.7 ± 0.0%         2.053        2.344
erubi-rails             1324.0 ± 0.0%   745.4 ± 0.0%         1.552        1.776
hexapdf                 2761.1 ± 0.0%  1300.2 ± 0.0%         1.545        2.124
liquid-c                  71.8 ± 0.0%    55.2 ± 0.0%         0.651        1.300
liquid-compile            56.4 ± 0.0%    41.3 ± 0.0%         0.604        1.366
liquid-il                 44.3 ± 0.0%    27.5 ± 0.0%         0.496        1.609
liquid-render            175.3 ± 0.0%    67.5 ± 0.0%         1.181        2.599
lobsters                1204.9 ± 0.0%  1133.9 ± 0.0%         0.635        1.063
mail                     167.1 ± 0.0%    83.0 ± 0.0%         0.616        2.013
psych-load              2414.1 ± 0.0%  1215.2 ± 0.0%         1.931        1.987
railsbench              2532.9 ± 0.0%  1457.8 ± 0.0%         1.256        1.738
rubocop                  231.2 ± 0.0%   318.8 ± 0.0%         0.374        0.725
ruby-lsp                 203.2 ± 0.0%   139.9 ± 0.0%         0.756        1.453
sequel                    55.6 ± 0.0%    42.1 ± 0.0%         1.365        1.322
shipit                  1954.0 ± 0.0%  1276.1 ± 0.0%         1.003        1.531
```

https://github.com/ruby/ruby-bench/actions/runs/29353153108/job/87154143699?pr=525

k0kubun (approved):

benchmarks/liquid-il/benchmark.rb
tekknolagi:
```suggestion
  1000.times do
```

eregon:
It would be quite interesting if this did the same workload as the [liquid-render benchmark](https://github.com/ruby/ruby-bench/blob/main/benchmarks/liquid-render/benchmark.rb), that way we could compare the relative performance of each rendering backend.

tekknolagi:
Good point!

PR #532, opened 2026-09-01
Carve out sudo commands for sudoers NOPASSWD

k0kubun:
This PR carves out the commands to modify CPU frequency and turbo boost as separate scripts so that we could let `/etc/sudoers` allow password-less sudo on them. It removes the need of password input without `--no-sudo`.


XrXr (commented):
I'm iffy on the security implications of this:

When one chooses to not add the commands to the sudoer file, we no long show the leaf commands that the script runs. You'd need to inspect the script file, which can change with a `git pull`.

If the commands in the sudoer file points to the script in the git repo, you have the same issue with the scripts updating with `git pull`. I think generally sudoer points to read-only binaries.

For allowing to run password-less, it looks like sudoer allows filtering by commandline arguments. Isn't it already possible to add each command ruby-bench runs to sudoer without changing ruby-bench?

PR #536, opened 2026-09-15
Add struct new micro benchmark

tekknolagi:
Benchmark allocating, writing, and reading fields


benchmarks/struct-new-no-escape.rb
```diff
@@ -0,0 +1,17 @@
...
+end
+
+run_benchmark(100) do
+  i = 0
+  while i < 1_000_000
+    test
+    i += 1
+  end
+end
+
```
k0kubun:
```suggestion
```

k0kubun (approved):
nice

benchmarks/struct-new-no-escape.rb
tekknolagi: augh. thanks

